PR #126, opened 2020-06-29
Reposilite is not starting up

massix:
Hello there,
first of all I'd wish to thank you for this project, I've been using it a lot lately for my projects and it's a great piece of software.

I'm using the Docker-based version on a K8S environment and I recently upgraded the docker image and noticed that it was no longer working, I think I might have found the issue, which is that the you probably forgot to run the Future.

It looks like it's working now on my side with this simple fix.

Would it be possible for you to merge this fix (or provide an equivalent fix if you have a better idea)?

Thanks a lot!


dzikoysk:
Yeah, I'm writing a lot of tests and it requires some architecture changes to integrate it well with JUnit. I forgot about it in the latest commit 😂 I've already fixed it, but I'll merge your commit to appreciate the contribution ❤️

Issue #138, opened 2020-07-09
UI/cosmetic improvements


jaqobb:
I suggest a few things regarding UI:
- add toggleable dark mode button (I'd say that this is pretty much just inverting/updating colors to be darker?). The reason I suggest that is I normally use Dark Reader to make websites darker (and I think a lot of people just prefer darker websites). However in case of Reposilite it doesn't really work. I guess theme for Reposilite can be somehow manually added to Dark Reader but it's always better to have an officially supported one (and I don't think it requires that much work),
- make website title modifiable. I believe that right now website title is unmodifiable and because of that it's pretty much the only way to recognize repository as being hosted using Reposilite (besides the website layout of course). What if the title becomes modifiable and Reposilite name was to be moved to the website footer (the bottom of the website isn't currently being used anyway). This way the important aspect of the website (its title) becomes modifiable and people are still gonna know what's being used to power up the repository. Since bottom of the website offers more space than the website's title, more information can be added there (like Reposilite's version). Besides that content can be made interactable (so let's say clicking on `Reposilite` will redirect either to the Reposilite's website or this repository).

dzikoysk:
The UI requires unified theme, but it's just not a priority for me at this moment. I'm pretty sure that time invested in functionalities has more value for this project, but I'm open for a 3rd party contributions.

jaqobb:
That's understandable. I also prefer adding new features over improving UI because that can pretty much wait. I don't really consider myself good at frontend but I'm gonna take a look at this issue later and if everything goes well I'll make a PR with my suggested changes regarding UI.

dzikoysk:
#460

PR #192, opened 2020-08-05
Add title customization and "powered by" footer

jolheiser:
This tackles the second item in #138 by providing a way to customize the header title and instead shows a "Powered by **Reposilite v1.2.3**" footer.

Since `title` is already used in the config, I opted for `websiteTitle`, as I'm not sure if there's a good way to "migrate" configs.

As well, it seems like the "production" compiled frontent is included in source...not sure if it should be but I committed it just in case. 🙂 

footer


jolheiser:
Sure, sounds good.

PR #196, opened 2020-08-05
Remove stats.yml from version control

### LINKED ISSUE #195: Exclude stats data file from workspace using .gitignore

jolheiser:
Didn't see an existing section in the `.gitignore` for Reposilite specific files, so I added it. Let me know if you'd rather put it elsewhere.

Resolves #195


dzikoysk:
I think it might be also a good idea to exclude these files:

image

Metadata files are generated dynamically, so there is no need to keep it in repository.

jolheiser:
@dzikoysk Done. Added it to the new section as well since the existing `Maven` section is more-so related to the project rather than the files the project produces...not sure if that makes sense. 

I can move it if you'd like. 🙂

Issue #259, opened 2020-09-20
Allow to increase request cache size in debug mode


dzikoysk:
Related to the conversation in #257 about large files, reported by @kay

PR #260, opened 2020-09-20
Added system property to set Javalin's requestCacheSize when using debug

kay:
Addresses https://github.com/dzikoysk/reposilite/issues/259 by way of system property.

* Only takes effect when debug is enabled
* Defaults to 8MB
* I've updated the FAQ but I wasn't sure how to test the links, I'm not much of a web dev, but I *think* the links should work


dzikoysk (approved):

dzikoysk:
Looks great, also thanks for adding it to the FAQ ❤️

PR #305, opened 2020-11-14
GH-306 Fix disk quota allocation when storing a file

oleludwig:
Fixes the bug, that the disk quota usage isn't updated correctly when uploading a new file. This is caused by reading the size of the locked file before the new file is written to it.
Adds a test to ensure that the disk quota usage is updated correctly.


dzikoysk:
Such a fair point, thanks for the PR :)

dzikoysk (approved):

PR #378, opened 2021-02-23
GH-379 Rewrite DeployService URLs

FormallyMyles:
This PR fixes an issue where if you deploy to a rewritten URL it doesn't work, e.g. http://localhost. The deploy does succeed but the files are written to the wrong directory (without the default repository.).

I've tested this and it works fine for us, but obviously, I might have missed something or there might be a better implementation.


dzikoysk (approved):

dzikoysk:
I didn't support this feature, because I thought I'll add an extra support for another build systems, but since I gave up on this idea, it's quite a good enhancement to the deployment. Thanks for PR! :)

PR #568, opened 2021-09-03
Clean up and simplify log.js

wvffle:


reposilite-frontend/src/store/log.js
```diff
@@ -1,78 +1,58 @@
...
+
 export default function useLog() {
-  levelNames.forEach(level => {
+  for (const level of levelNames) {
     levels[level] = {
       name: level,
       enabled: true,
-      count: 0,
-      test: (message) => message.includes(`&nbsp;|&nbsp;${level.toUpperCase()}&nbsp;|&nbsp;`)
+      count: computed(() => rawLog.reduce((accumulator, entry) => accumulator + (entry.level === level), 0))
```
wvffle, line 32: You can optimize this by making a global computed like:
```js
const count = computed(() => rawLog.reduce((accumulator, entry) => {
  accumulator[entry.level] ?= 0
  accumulator[entry.level] += 1
  return accumulator
}, {}))
```

And using: 
```
count: count[level]
```

wvffle: The `variable ?= value` syntax is setting the value of `variable` to  value of `value` only if the current `variable` value is nullish. In case you didn't know.

reposilite-frontend/src/components/Console.vue
```diff
@@ -96,8 +87,8 @@ export default {
     }
 
     watch(
-      () => active.value,
-      isActive => isActive ? setupConnection() : close(),
+      () => selectedTab.value,
```
wvffle, line 90: This actually can be written as `watch(selectedTab, ...)`

dzikoysk (approved):

Issue #558, opened 2021-08-29
Move away from openjdk alpine docker image to be able to deploy on arm64 systems


Minecraftschurli:
the openjdk:<version>-alpine images only support amd64 which makes it impossible to deploy reposilite on arm64 systems (like a raspberrypi 4) so it would be nice if you could replace it with another image (like openjdk:<version>-slim)

dzikoysk:
Tbh I don't know the difference between alpine and slim releases, I use alpine as it's recommended image. I've seen that ARM64 users in fact uses the slim release:

* https://github.com/dzikoysk/reposilite/issues/516#issuecomment-859168689

Does it delivers the same packages for every platform?

Minecraftschurli:
well the slim images are based on debian so it uses the apt package manager instead of alpines apk but there should not be a difference between arm64 and amd64

dzikoysk:
Do you want to make a PR? I think we could apply it to the `master` and `3.0` branches. If not, I'll take care of it during the release phase of 3.x version.

Minecraftschurli:
I tried it on the master branch and it failed with an error in while running the tests. I am currently trying it on 3.0

dzikoysk:
Master branch contains latest stable version of 2.x and it shouldn't fail. Could you try to paste the tests with errors? 🤔 Speaking of the 3.0, it's obviously still not finished, so the final jar won't work yet anyway.

Minecraftschurli:
This is the log of the error when building 2.9.24 on arm64
```log
#15 362.5 [INFO]
#15 362.5 [INFO] -------------------------------------------------------
#15 362.5 [INFO]  T E S T S
#15 362.5 [INFO] -------------------------------------------------------
#15 363.1 [INFO]
#15 363.1 [INFO] Results:
#15 363.1 [INFO]
#15 363.1 [INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
#15 363.1 [INFO]
#15 363.2 [INFO] ------------------------------------------------------------------------
#15 363.2 [INFO] Reactor Summary for reposilite-parent 2.9.24:
#15 363.2 [INFO]
#15 363.2 [INFO] reposilite-parent .................................. SUCCESS [ 43.425 s]
#15 363.2 [INFO] reposilite-backend ................................. FAILURE [05:01 min]
#15 363.2 [INFO] ------------------------------------------------------------------------
#15 363.2 [INFO] BUILD FAILURE
[...]

PR #589, opened 2021-09-06
GH-558 Enable ARM64 platform target

dzikoysk:
Dirty branch that combines various attempts to launch Gradle build on ARM64 platform.


dzikoysk:
Ok, I think it just impossible to launch this Gradle setup on arm64 xD

Issue #458, opened 2021-04-22
Reposilite 3.0


dzikoysk:
Assumptions:
* Keep resource consumption as low as possible
* Keep the service as simple as possible
* Provide a convenient migration option for 2.x users
* Improve the scalability
* Cleanup legacy sources

Pull request: [3.0](https://github.com/dzikoysk/reposilite/pull/465)
Progress: [Reposilite 3.0](https://github.com/dzikoysk/reposilite/projects/2)
~ [3.0 Milestone](https://github.com/dzikoysk/reposilite/milestone/3)
~ [3.0 Branch](https://github.com/dzikoysk/reposilite/tree/3.0)

**Note:** If you're new user, you can use latest 2.9.x releases. It's stable and fully documented version of Reposilite.

PR #659, opened 2021-09-21
GH-458 Added repository name to maven id

d0x7:
Added this to my personal build and thought it could be a good addition for the main release.

When on the overview and selecting a repository, e.g. releases, it adds "-releases" behind the repository id. 
Nothing is added when no repository is selected (unlike the domain, where "{repository}" is added instead. I don't think this would be good here, so I just left it as is).

image


dzikoysk:
Thanks for contribution!

It might be a good idea, because in most cases people use different id for releases/snapshots to provide different credentials/metadata associated with the given repository. 

By the way, keep in mind that 3.0.0-alpha.1 is still experimental and may feel buggy until the official release. If you're looking for stable solution, 2.x might be still a better choice for production env :)

Issue #725, opened 2021-10-01
Use shadowJar result explicitly in Dockerfile


dzikoysk:
Use JAR with `-all.jar` suffix

PR #736, opened 2021-10-01
GH-725 Copying shadowJar result in Dockerfile.

shitzuu:
This pull request solves #725


dzikoysk:
I forgot that this file has a little different name, so it cannot be `-all.jar` in the end:

image

🤔

dzikoysk (approved):

dzikoysk:
Thanks!

PR #740, opened 2021-10-01
GH-668 Reimplement card

### LINKED ISSUE #668: Reimplement Card component
**Component**

* https://github.com/dzikoysk/reposilite/blob/main/reposilite-frontend/src/components/browser/Card.vue

**Context**

In various situations `Card.vue` component behaves differently.

cc: @wvffle

wvffle:
Fix #668

### DIFF
--- a/reposilite-frontend/src/components/browser/Card.vue
+++ b/reposilite-frontend/src/components/browser/Card.vue
@@ -154,6 +154,12 @@ export default {
   border: transparent;
   margin-top: 10px;
 }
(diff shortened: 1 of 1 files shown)


wvffle:
The downside of this implementation is that I use only one element as snippet content container, so to animate it I need to first animate it out, then change content and then animate it in. That's why I've added the opacity to the animation.

dzikoysk (approved):

dzikoysk:
Thanks once again! :)

PR #747, opened 2021-10-03
GH-745 Focus console

### LINKED ISSUE #745: Focus input field during entering the Console view
### Request details

^

jolheiser:
Closes #745 

To be clear before reading further, this PR **does** focus the input _after_ the transition occurs.  
Below is just an explanation on why I deferred to `setTimeout`.

Please let me know if you are aware of a cleaner solution.

-----

When focusing the console immediately (even using `nextTick`), I ended up with a UI glitch like so

Screenshot from 2021-10-03 14-45-35

Setting the timeout was the only way I could get the animation to finish before focusing. It's more or less imperceptible, but I tend to dislike "timing" solutions.

I was unable to hook into the events from the tab library used in a way that didn't cause the above glitch, either.

### DIFF
--- a/reposilite-frontend/src/components/Console.vue
+++ b/reposilite-frontend/src/components/Console.vue
@@ -35,6 +35,7 @@
       </div>
       <hr class="dark:border-dark-300">
       <input
+        id="consoleInput"
         placeholder="Type command or '?' to get help"
         class="w-full py-2 px-4 rounded-b-lg bg-white dark:bg-gray-900 dark:text-white"
         v-model="command"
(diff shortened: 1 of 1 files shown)


dzikoysk:
Thanks for PR! :) The [tab library](https://github.com/HJ29/vue3-tabs) I use is kinda... rough and doesn't work well with various other components. I also don't really like time-based solutions, but I think it should be enough for now - in case of a better solution we can always modify this in the future.

dzikoysk (approved):

PR #748, opened 2021-10-04
Fix Filter

### LINKED ISSUE #744: Channel filters does not match messages in production env
### What happened?

All messages are counted as `Other`

image
image

### Reposilite version

3.x

### Relevant log output

_No response_

jolheiser:
Fixes #744 

If you look at the screenshot in #744 you'll notice the Java/Kotlin logger only sends a single pipe (`|`).  
This is one solution, another would be to instead insert the required extra pipe in the Java/Kotlin logger (whereas this changes the search to only look for one).

I've also changed the fake API logger to only send a single pipe for closer consistency.

### DIFF
--- a/reposilite-frontend/fake-api/extensions.js
+++ b/reposilite-frontend/fake-api/extensions.js
@@ -35,7 +35,7 @@ const invalidCredentials = (res) =>
   res.status(401).send('Invalid credentials')
 
 const sendMessage = (connection, message) =>
-  connection.send(new Date().toDateString() + " | " + message)
+  connection.send(`${new Date().toDateString()} ${message}`)
 
 const createFileDetails = (name, contentType, contentLength) =>
   ({ type: 'FILE', name, contentType, contentLength })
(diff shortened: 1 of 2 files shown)


jolheiser:
I should note, I was unable to get this compiled to do a _true_ test, unfortunately Kotlin seemed to keep complaining about Kapt and I was unable to resolve it properly.

If you would feel more confident about this PR I can try more tomorrow, but this seemed like a fairly straightforward change. :slightly_smiling_face:

dzikoysk (approved):

dzikoysk:
I didn't notice it to be honest 😅 

> I should note, I was unable to get this compiled to do a _true_ test, unfortunately Kotlin seemed to keep complaining about Kapt and I was unable to resolve it properly.

Reposilite currently cannot be compiled on JDK17, because Kapt does not support this yet (#587). It barely support JDK16, because newer versions of Kotlin (1.5.31+) breaks the compatibility. What was your environment? 🤔 

Thanks for PR! :)

Issue #743, opened 2021-10-02
 Responsive UI


dzikoysk:
### Request details

Well...

![Screenshot_2021-10-02-03-30-35-52](https://user-images.githubusercontent.com/4235722/135699683-7ffdd7bc-a2ac-4a55-9b8a-d62808cfb25c.jpg)

sudhinm:
can I pick this?

dzikoysk:
Feel free to contribute! :) This issue addresses a few problems, so remember that we prefer small changes in multiple standalone PRs rather than one big PR.

dzikoysk:
Another issue:

![image](https://user-images.githubusercontent.com/4235722/135926917-8aa8cd05-0f9e-47eb-8664-a5148cbc065a.png)

sudhinm:
Cool..got it..thanks

dzikoysk:
I've found that we're missing a font in code snippets on your platform. 

![Screenshot_2021-10-08-09-21-14-28](https://user-images.githubusercontent.com/4235722/136514863-a3413f87-8f78-447b-ad0b-dc6fc1c931db.jpg)

It should be much more smooth like on the screen.

PR #773, opened 2021-10-05
GH-743 Fix hero content alignment and buttons in the mobile view

sudhinm:
This fixes the hero content alignment in the mobile view. Sign in and theme switcher button shrinking issue also have fixed.

Before
Screenshot 2021-10-05 at 11 35 41 PM

After
Screenshot 2021-10-05 at 11 36 17 PM

Issue - #743


dzikoysk:
I think that padding below and above Hero component could be a little smaller and padding below image could be a little increased, what do you think about it?

sudhinm:
> I think that padding below and above Hero component could be a little smaller. Also, padding below image could be a little increased. What do you think about it?

Sure, that would look better.

sudhinm:
Is this ok @dzikoysk ?
Screenshot 2021-10-06 at 12 18 14 AM

dzikoysk (approved):

dzikoysk:
Yeah, it's better now 👍

Issue #673, opened 2021-09-22
`Repository details` are shown when in specific artifact directory.


wvffle:
![image](https://user-images.githubusercontent.com/13330620/134429602-69d6c8bd-9849-4b78-8f43-cc176409d270.png)
![image](https://user-images.githubusercontent.com/13330620/134429614-eb2afb9f-f007-4e09-979b-2ba8141d0ae4.png)

PR #774, opened 2021-10-06
show artifact details in artifact directory

PyrrhaDevs:
#673 

image
image

Sorry about the unverified second commit, I forgot to set my signing key in GitKraken.


dzikoysk:
Thanks! In the future we may consider some kind of caching to improve response times and ui avoid flickering, but it's fine for now to just provide valid data for user

Issue #695, opened 2021-09-26
Support token management through API endpoints


dzikoysk:
### Request details

Enhance 'token' domain with the given public API:

```shell
GET    /api/tokens         # Get all tokens (manager only)
GET    /api/tokens/{name}  # Get token details (this token & manager)
PUT    /api/tokens/{name}  # Update or create token (manager)
DELETE /api/tokens/{name}  # Delete token (manager)
```

PR #775, opened 2021-10-06
GH-695 Implement an access token API to manage tokens

ChachyDev:
#695


reposilite-backend/src/main/kotlin/com/reposilite/token/AccessTokenFacade.kt
```diff
@@ -17,24 +17,36 @@ package com.reposilite.token
 
 import com.reposilite.token.AccessTokenSecurityProvider.B_CRYPT_TOKENS_ENCODER
 import com.reposilite.token.AccessTokenSecurityProvider.generateSecret
-import com.reposilite.token.api.AccessToken
-import com.reposilite.token.api.AccessTokenPermission
-import com.reposilite.token.api.AccessTokenType
+import com.reposilite.token.api.*
```
dzikoysk, line 20: We shouldn't use wildcard imports at all, especially in Kotlin to avoid strange issues in some specific cases and keep all references clean :)

reposilite-backend/src/main/kotlin/com/reposilite/token/api/TokensResponse.kt
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package com.reposilite.token.api
+
+data class TokensResponse(val tokens: List<AccessTokenField>) {
+    data class AccessTokenField(
```
dzikoysk, line 21: Frontend and other components use just deserialized AccessToken, so there is no point to create dedicated dto at this moment for this, so just use them directly

reposilite-backend/src/main/kotlin/com/reposilite/token/api/AccessToken.kt
```diff
@@ -68,9 +68,16 @@ enum class AccessTokenPermission(val identifier: String, val shortcut: String) {
         fun findAccessTokenPermissionByIdentifier(identifier: String): AccessTokenPermission =
             values().first { it.identifier == identifier }
 
+        private fun findAccessTokenPermissionByIdentifierNullable(identifier: String): AccessTokenPermission? =
```
dzikoysk, line 71: Might be a good idea to just improve these old methods rather than introducing the new one with `Nullable` suffix. In sources that required non-null values just use `!!` / provide an error

reposilite-backend/src/main/kotlin/com/reposilite/token/AccessTokenFacade.kt
```diff
@@ -70,4 +82,49 @@ class AccessTokenFacade internal constructor(
...
+        if (!manager) {
+            return errorResponse(HttpCode.UNAUTHORIZED, "You must be a manager to access this endpoint!")
+        }
+
+        if (getToken(request.name) != null) {
+            deleteToken(request.name)
+        }
+
+        return createAccessToken(request).asSuccess()
+    }
```
dzikoysk, line 129: A few notes related in general to these methods:

1. In general a facade should not be responsible for authorization, so rather than passing the `manager: Boolean` property, verify this in endpoints. 
2. `getTokenResponse` methods should be called more like just `getTokens`/`findTokens`
3. Passing `tokens: Collection<AccessToken>` looks kinda dirty, I think you can just take it from facade inside this metod

reposilite-backend/src/main/kotlin/com/reposilite/token/api/CreateAccessToken.kt
```diff
@@ -22,6 +22,11 @@ data class CreateAccessTokenRequest(
     val permissions: Set<AccessTokenPermission> = emptySet()
 )
 
+data class CreateAccessTokenRequestWithNoName(
```
dzikoysk, line 25: I think `CreateAccessTokenWithNoNameRequest` might be a better name, what do you think?

dzikoysk (changes requested):

dzikoysk:
Thanks for PR! Overall it looks quite good, but some things could be improved to simplify it - for now I highlighted the most important. in case of any problems feel free to ask :)

reposilite-backend/src/main/kotlin/com/reposilite/token/AccessTokenFacade.kt
ChachyDev: Alright, believe Intellij automatically did this so I'll quickly revert it

reposilite-backend/src/main/kotlin/com/reposilite/token/api/TokensResponse.kt
ChachyDev: Hmmm, I decided to use this instead because I didn't think it would be the best idea to return the secret also but if you believe that's alright then I'll switch it to that

reposilite-backend/src/main/kotlin/com/reposilite/token/application/AccessTokenWebConfiguration.kt
```diff
@@ -55,4 +49,5 @@ internal object AccessTokenWebConfiguration : WebConfiguration {
         consoleFacade.registerCommand(RouteRemove(accessTokenFacade))
     }
 
+    override fun routing(reposilite: Reposilite) = setOf(AccessTokenApiEndpoints(reposilite.accessTokenFacade))
```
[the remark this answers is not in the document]
ChachyDev: By new, I assume a new line? If so 👍

reposilite-backend/src/main/kotlin/com/reposilite/token/api/TokensResponse.kt
ChachyDev: Never mind, I was mistaken! Fixing and marking as resolved

dzikoysk:
Good job! Some cosmetic issues, but I'll take care of it later as I don't really require from you to follow my codestyle 1:1, especially that's it's not precisely defined.

Might be a good idea to cover this api with unit & integration tests, do you want to try to do this in a new PR or you're done for now? :)

ChachyDev:
> Good job! Some cosmetic issues, but I'll take care of it later as I don't really require from you to follow my codestyle 1:1, especially that's it's not precisely defined.
> 
> Might be a good idea to cover this api with unit & integration tests, do you want to try to do this in a new PR or you're done for now? :)

I'll start on the unit & integration tests now

PR #779, opened 2021-10-06
Issue-#743 Fix levels count overflow issue in the console view

sudhinm:
Issue #743 

Desktop View
Screenshot 2021-10-07 at 12 01 46 AM

iPad view
Screenshot 2021-10-07 at 12 02 24 AM

Mobile view
Screenshot 2021-10-07 at 12 02 44 AM


sudhinm:
> Thanks again :) Does it work well with up to 4 digits in count number and on a dark theme?

Yes, It will work on the dark mode too. Also it won't break the line even for 4+ digits. 


Screenshot 2021-10-07 at 10 08 23 AM

PR #787, opened 2021-10-08
GH-743 Add fallback mono font for MacOS

sudhinm:
Added 'Monaco' font as a fallback for 'Consolas' font.
#743 

Screenshot 2021-10-08 at 4 25 19 PM

Screenshot 2021-10-08 at 4 24 56 PM


dzikoysk:
Thanks! :)

PR #823, opened 2021-10-15
Removed duplicate signin call

WithoutAName25:
signin was called twice if you clicked on the button and didn't pressed enter


dzikoysk:
Thanks! I didn't notice it 😳

Issue #826, opened 2021-10-18
Support hot-reloading of configuration


dzikoysk:
### Request details

It would be great if we could reconfigure Reposilite instance without a need to restart the whole service. So on, it might be a good enhancement towards the configuration panel in the dashboard.

PR #839, opened 2021-10-19
GH-826 Support hot-reloading of configuration

dzikoysk:
To do:
* Improved configuration handling
* Shared configuration state between instances


dzikoysk:
Requires support for references in CDN:

* https://github.com/dzikoysk/cdn/issues/62

PR #867, opened 2021-10-26
GH-556 Implement allowed groups for remote repositories

### LINKED ISSUE #556: Proxy store whitelist
I would like to have a whitelist (per-package/per-root-package) to only allow some specific proxied artifacts.

For example:
```
proxyWhitelist: [
  "a.simple.package.*" // Any artifact that starts with "a.simple.package"
  "this.specific.package"
]
```

So I can whitelist the packages that I want and prevent other users to use my repository to cache them all.

Saladoc:
If no --allow parameters are specified, any artifacts are considered allowed. As soon as one allowed group is given, only artifacts that are prefixed with one of the allowed groups will be requested from the repository.

I've been trying to include a unit test but haven't successfully investigated how you do your testing so I postponed it to first get input on the solution.

Closes #556


dzikoysk (approved):

Saladoc:
I've now added the tests as well. Looks like I had some wrong assumptions on the contents of the gav variable in the ProxyService

dzikoysk (approved):

dzikoysk:
Thanks!

PR #936, opened 2021-11-22
GH-937 Update S3 client

envomp:
S3 client change accessKey, secretKey, region to optional parameters, so users can use default configuration provided by AWS
S3 storage provider create bucket if it's missing only


dzikoysk (approved):

dzikoysk:
Thanks for your contribution! ❤️

Issue #956, opened 2021-12-04
Allow Maven Badges to be filtered


hypherionmc:
### Request details

It would be useful to have a way to "filter" artifacts when using the badges api.

Say I have a deployment with multiple versions (like a minecraft mod for example) with the following version numbers:

1.16.5-1.2
1.17.1-2.1
1.18-1.0

then the filter would allow you to show the latest artifact starting with '1.16.5' or '1.17.1'

PR #980, opened 2021-12-13
GH-956 allow versions to be filtered

WithoutAName25:
If you have something that should be implemented differently just write me. It might be also useful to add a filter to show only `alpha`, `beta` or `release` versions, but I think that should be in a second pull request if we want to implement this.

### DIFF
--- a/reposilite-backend/src/main/kotlin/com/reposilite/badge/BadgeFacade.kt
+++ b/reposilite-backend/src/main/kotlin/com/reposilite/badge/BadgeFacade.kt
@@ -2,7 +2,7 @@ package com.reposilite.badge
 
 import com.reposilite.badge.api.LatestBadgeRequest
 import com.reposilite.maven.MavenFacade
-import com.reposilite.maven.api.LookupRequest
+import com.reposilite.maven.api.VersionLookupRequest
 import com.reposilite.web.http.ErrorResponse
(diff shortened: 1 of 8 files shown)


dzikoysk:
Thanks for PR! ❤️ 

I think that if we want to provide filter for prefix & suffix, we should implement this here as it's a part of API and we should not modify it in the future if possible. What do your think about regex? 🤔

WithoutAName25:
I thought regex would be to difficult to use and it would be easier to do sth like `?filter=2.0&type=release` to get all versions that start with `2.0` and are releases and not betas or alphas.
And with regex you maybe have to encode / escape some special characters. But if we implement a feature in the frontend where you can generate badge urls (at some point in the future), regex might be the better option.
But it would be no problem to switch to regex, I would have to edit just one line of code.

WithoutAName25:
As reference:
https://shields.io/maven-metadata/v?metadataUrl=https%3A%2F%2Frepo1.maven.org%2Fmaven2%2Fcom%2Fgoogle%2Fguava%2Fguava%2Fmaven-metadata.xml&versionPrefix=29.&versionSuffix=-android
That's how shields.io implemented this

dzikoysk (approved):

dzikoysk:
Makes sense, the PR also looks good! Let's merge it.

PR #981, opened 2021-12-18
GH-555 Support plugins

dzikoysk:


dzikoysk:
Todo:
- Topological sort algorithm impl to resolve dependnecies 
- `ReposilitePlugin` instance loader
- File based list of plugins in JAR
  - In the future: Annotation processor
- JAR loader
- Event based registration of commands

dzikoysk:
TODO: Use parent gradle wrapper instead of the old `./reposilite-backend` one

Issue #1011, opened 2021-12-30
Support PAC4J security engine


dzikoysk:
### Request details

https://www.pac4j.org:

* https://github.com/dzikoysk/reposilite/issues/924
* https://github.com/dzikoysk/reposilite/issues/1009

EthanDevelops:
https://github.com/pac4j/javalin-pac4j

dzikoysk:
Ok, so I've made some research about it and I see some issues related to this topic. 

The first one is just the type of authentication form. Currently, Reposilite (and pretty much all microservice-like apps) uses stateless API to communicate between client (e.g. frontend, maven, external apps) and server (one or more instances).

![stateless](https://user-images.githubusercontent.com/4235722/150143228-fdbe763f-27eb-448e-a29d-effb3509a88e.png)

In pac4j it's pretty much represented by direct clients and there are authenticators that works this way:

![authenticators](https://user-images.githubusercontent.com/4235722/150143624-ff0bd898-436a-471f-8930-9c0c1413694a.png)

Speaking of OAuth based services, OIDC and in general all indirect clients:

![session](https://user-images.githubusercontent.com/4235722/150143882-ed52af6e-f9b6-4030-a592-ed87af94439e.png)

The indirect mechanism requires session to store authentication related stuff secured:

![session clients](https://user-images.githubusercontent.com/4235722/150144111-cad9e900-af1e-47db-92a5-634b53d55a84.png)

Because we can't really use session, that's just not an option for us. Of course, the OAuth implementation can expose JWT token and it could be up to the client to properly handle it, but to be honest, what's the benefit of this feature anyway then?

What's the real scenario of using alternative auth method, despite of easier logging to web ui, I guess there is no real benefit of it as it doesn't really work with Maven itself. 🤔

pauliesnug:
@dzikoysk You could try looking into https://docs.gradle.org/current/userguide/declaring_repositories.html#sub:http-header-auth, since this would allow for PAC4J tokens and credentials to be passed as HTTP headers. This could allow for a custom Gradle plugin that allows for OAuth2-handled Maven publishing!

dzikoysk:
When we were talking about this feature, Reposilite only had access tokens. In 4.x I'm going to add a regular user system, so it makes sense to provide OAuth - it'll be easier to manage people in the organization. In such case, each user would generate their access tokens in the UI, so I guess it's no longer an issue 😅

pauliesnug:
[...]

PR #1014, opened 2021-12-30
GH-1011 Add Pac4J auth.

EthanDevelops:
#1011


dzikoysk:
Btw I'm not sure if we want to use this Javalin extension, it's mostly about authenticating request using `before` handler and we should probably handle it through `AuthenticationFacade` manually to avoid this kind of `before` -> `get` flow 🤔

EthanDevelops:
We could adapt the core library to the current auth system.

EthanDevelops:
Tags request:
dependencies, security, and java

dzikoysk:
In most cases we don't tag PRs, it's the responsibility of issue to properly address conversation. I'll add security tag tho :)

dzikoysk:
#1032 can affect this also

dzikoysk:
Hey @EthanDevelops, what's the status of this PR? :)

EthanDevelops:
Hey! It’s a bit stale at the moment because I think that the current auth implementation would be pretty hard to make work with Pac4j. Maybe the auth for the website could use some sort of OAuth system that would be verified in the backend. Maybe an API key solution could work?

dzikoysk:
Not sure how it works yet, I guess that this feature might be breaking as I thought, so we need to take care of this before release of stable 3.x. I'll try to read some pac4j docs this week.

Issue #993, opened 2021-12-23
JavaDoc Hosting


TOTHT0MI:
### Request details

(I've searched a lot but couldn't find any requests related to this topic, so here I am, but sorry if it's a duplicate)

My **idea/request** is simple:

Host JavaDoc pages from the maven repo -javadoc.jar file as a static page in an iframe.
The iframe would be embedded in the main site on a separate route. (For security the headers would only allow iframes from the same-origin)
This feature would also introduce a separate badge that would show the version of the latest JavaDoc available for a specified artifact or N/A if there are no available ones.

If it detects that the published artifact has a Javadoc generated along with it, with RegExp for example; the backend could extract the HTML content with the following:
The -javadoc.jar file is actually just like a "zip", it only contains the HTML pages. By running the command: `jar xf *-javadoc.jar`, it will extract the jar's content into the folder it's in. (* would be the artifact name).

(P.S: I'd create a pull request and implement it myself, but I don't have much experience with Kotlin, Javalin, and especially with your codebase. Once I learn it - will probably do - I most likely come back and implement it myself, if someone else hasn't implemented it by that time)

dzikoysk:
Interesting! I think it could be great to somehow host these pages, but I don't think that the iframe concept is a good idea 🤔 Imo we should support this though standalone route repo.domain.com/javadocs/<artifact url>, it looks better and we can host it as a standard page (such as e.g. /swagger)

dzikoysk:
As I see you've forked Reposilite, keep in mind that #981 is quite breaking as it refactors a lot of things. I'd suggest to wait a little bit, because I'm about to finish it soon :)

TOTHT0MI:
Sure, thanks for notifying me!
I'm still going over how things are structured and interact with each other, and will only experiment at the moment, coming up with how to do it most efficiently. :)

So that breaking refactor will not be a problem, as I would implement it from scratch free of junk code after experimentation and after I've settled on a working solution.

dzikoysk:
Sure, let me know about confusing parts, I'd like to prepare quick page for people that don't know Kotlin to simplify contribution process. I'm quite sure that with such a post people that know Java should be able to learn Kotlin in about 5mins, as it works pretty much the same, it's all about this extra syntax sugar :)

dzikoysk:
>Will probably use the event system to listen for deploys and handle the JavaDoc extraction there.
Maybe move this entire thing into a plugin as an optional addon.

[...]

Issue #927, opened 2021-11-18
Replace Path based API & operations with abstract VirtualResource layer


dzikoysk:
### Request details

Benefits:
* Improves paths handling 
* Non platform dependent implementation
* Improves tracing and monitoring
* Corresponds better to the usage with various StorageProviders

PR #1017, opened 2022-01-02
GH-993 Initial commit of JavaDoc plugin

TOTHT0MI:
Initial commit of JavaDoc plugin prototype (functional), described in #993 :

- Added route /javadoc/repo/gav (at the moment gav must point to a javadoc.jar file directly: see TODO in JavadocRoute.kt)
- Created JavaDoc extractor (extracts the files from a javadoc.jar file either via command-line or programmatically; extracted docs are stored in workingdir/javadocs/repo/stripped-gav)
- JavaDoc is embedded in a index file using an iframe described in JavadocRoute.kt (this is so we can have buttons on top like: download javadoc, etc.)

Missing features:
- Handling of /latest & non direct gavs (ex.: com/group/artifact/version/index.html)
- Access Token handling
- Ability to download javadocs in a zip format (optional)


TOTHT0MI:
TODO: Remove shell command based extraction.

dzikoysk:
Changes on `main` related to #927 & #1018 may affect this PR, just fyi

TOTHT0MI:
IMO as an alpha release plugin should be done, there are many things that could be implemented in the future, tried to leave as much space as possible for those.

dzikoysk:
Could you try to pull out the logic from `JavadocRoute` (that probably should be called `JavadocEndpoints`) and move it to `JavadocFacade`? It's just a cleaner solution to separate reuqest handling and business logic, so we can's also provide unit & integration tests in the future :)

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/DocExtractor.kt
```diff
@@ -0,0 +1,51 @@
...
+import java.io.File
+import java.io.IOException
+import java.nio.file.Path
+import java.util.jar.JarFile
+
+/**
+ * @author TOTHTOMI
+ * @version 1.0.0
+ * @since Dec. 27, 2021
+ */
```
dzikoysk, line 12: We don't really use authority comments, it's hard to maintain and doesn't provide valid data for long living projects - authors are listed in git history per each file :)

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/DocExtractor.kt
```diff
@@ -0,0 +1,51 @@
...
+internal class DocExtractor {
+
+    @Throws(Exception::class)
```
dzikoysk, line 15: Error handling through exceptions is pretty much antipattern, sth like `Result<Unit, ErrorResponse/Exception>` might be a better choice

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocPlugin.kt
```diff
@@ -0,0 +1,39 @@
...
+class JavadocPlugin : ReposilitePlugin() {
+
+    private lateinit var reposilite: Reposilite
+    private lateinit var javadocFolder: File
+
+    override fun initialize(): Facade? {
+        extensions().registerEvent(ReposiliteInitializeEvent::class.java) { event: ReposiliteInitializeEvent ->
+            logger.info("")
+            logger.info("--- javadoc plugin")
+            logger.info("JavaDoc plugin has been successfully loaded!")
```
dzikoysk, line 22: If plugin does not print any relevant data during startup, I think there is no point to add extra things like that here 🤔

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocPlugin.kt
```diff
@@ -0,0 +1,39 @@
...
+
+            reposilite = event.reposilite
+            javadocFolder = File(reposilite.parameters.workingDirectory.toFile().absolutePath + File.separator + "javadocs")
```
dzikoysk, line 25: I think you can just use `reposilite.parameters.workingDirectory.resolve("javadocs")` and then use methods from `Files` class to operate on files using a new NIO api :)

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocPlugin.kt
```diff
@@ -0,0 +1,39 @@
...
+import java.io.File
+
+@Plugin(name = "JavadocPlugin")
+class JavadocPlugin : ReposilitePlugin() {
+
+    private lateinit var reposilite: Reposilite
+    private lateinit var javadocFolder: File
+
+    override fun initialize(): Facade? {
+        extensions().registerEvent(ReposiliteInitializeEvent::class.java) { event: ReposiliteInitializeEvent ->
```
dzikoysk, line 19: You can listen for events directly through `ReposilitePlugin#event` method, e.g.:

```kotlin
event { event: RoutingSetupEvent ->
    event.registerRoutes(MavenEndpoints(mavenFacade, frontendFacade, settingsFacade))
    event.registerRoutes(MavenApiEndpoints(mavenFacade))
    event.registerRoutes(MavenLatestApiEndpoints(mavenFacade, settingsFacade))
}
```

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocPlugin.kt
dzikoysk: Also, `parameters` are available in `extensions().parameters`, so you don't really need `Reposilite` instance here

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocPlugin.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.reposilite.plugin.api.ReposiliteInitializeEvent
+import com.reposilite.plugin.api.ReposilitePlugin
+import com.reposilite.web.api.RoutingSetupEvent
+import java.io.File
+
+@Plugin(name = "JavadocPlugin")
+class JavadocPlugin : ReposilitePlugin() {
+
+    private lateinit var reposilite: Reposilite
+    private lateinit var javadocFolder: File
```
dzikoysk, line 16: Keeping these values as lateinit fields is kinda bad practice and there is no point to do this, check next comments below.

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocRoute.kt
```diff
@@ -0,0 +1,237 @@
...
+                        val source = extractJavaDoc(newGav, repo, mavenFacade, javadocFolder, this)
+
+                        ctx.encoding(Charsets.UTF_8)
+                        ctx.contentType(ContentType.TEXT_HTML)
+                        response = source
+                        return@consume
+                    }
+
+                    // TODO handle latest
+                    println("hi")
```
dzikoysk, line 74: 😳

dzikoysk (changes requested):

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/DocExtractor.kt
TOTHT0MI: Ok, that's autogenerated by my Intellij, but I will remove it.

TOTHT0MI:
I have some problems after introducing JavadocFacade:

```
Supertypes of the following classes cannot be resolved. Please make sure you have the required dependencies in the classpath:
    class com.reposilite.token.api.AccessToken, unresolved supertypes: net.dzikoysk.exposed.shared.IdentifiableEntity
```
And:
```
Inheritance from an interface with '@JvmDefault' members is only allowed with -Xjvm-default option
```

Currently the constructor looks like this:
```kotlin
class JavadocFacade internal constructor(
    private val javadocFolder: Path,
    private val mavenFacade: MavenFacade,
    private val journalist: Journalist
) : Journalist, Facade {
```

And I used the access token like this:
```kotlin
fun resolveRequest(repo: String, gav: String, extension: String, accessToken: AccessToken?): Result<JavadocResponse, ErrorResponse> {
    [...]
    val lookUp = mavenFacade.findDetails(LookupRequest(accessToken, repo, newGav));
}
```
I haven't changed the gradle build script.

dzikoysk:
For jvm default:

* https://github.com/dzikoysk/reposilite/blob/087e97fd87a2f9c77640f5a69c0e5fa5e845e632/reposilite-backend/build.gradle.kts#L149

For IdentifiableEntity, use this dependency with compileOnly:

* https://github.com/dzikoysk/reposilite/blob/087e97fd87a2f9c77640f5a69c0e5fa5e845e632/reposilite-backend/build.gradle.kts#L66

reposilite-plugins/javadoc-plugin/build.gradle.kts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+application {
+    mainClass.set("com.reposilite.plugin.javadoc.JavadocPluginKt")
+}
+
+dependencies {
+    // compileOnly("org.panda-lang:reposilite:version") for external plugins
+
+    implementation("net.dzikoysk:exposed-upsert:1.0.3")
```
dzikoysk, line 17: Should be `compileOnly`, because `implementation` + `shadow` plugin will bundle dependency in plugin jar and we don't want it to avoid conflict between classes of the same dependency at runtime

reposilite-plugins/javadoc-plugin/build.gradle.kts
```diff
@@ -0,0 +1,31 @@
...
+plugins {
+    kotlin("jvm")
+}
+
+application {
+    mainClass.set("com.reposilite.plugin.javadoc.JavadocPluginKt")
+}
+
+dependencies {
+    // compileOnly("org.panda-lang:reposilite:version") for external plugins
```
dzikoysk, line 15: Unnecessary comment

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/DocExtractor.kt
```diff
@@ -0,0 +1,56 @@
...
+import com.reposilite.web.http.errorResponse
+import io.javalin.http.HttpCode
+import panda.std.Result
+import panda.std.Unit
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.Path
+import java.util.jar.JarFile
+
+internal class DocExtractor {
```
dzikoysk, line 13: Shouldn't this class be a part of JavadocFacade?

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/DocExtractor.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    fun extractJavadoc(jarFilePath: Path, destination: Path): Result<Unit, ErrorResponse> {
+        val file = jarFilePath.toFile()
+        val dest = destination.toFile()
+
+        // Some checks, to make sure we're working with valid files/paths.
+        require(!file.isDirectory) { "JavaDoc jar file must not be a directory!" }
+        require(dest.isDirectory) { "Destination must be a directory!" }
+        check(file.name.contains("doc.jar")) { "Invalid javadoc.jar! Name must contain: \"doc.jar\"" }
```
dzikoysk, line 22: These methods may throw exception + there is no need to use `File` instead of Path, e.g.:
```kotlin
if (Files.isDirectory(jarFilePath) return errorResponse(BAD_REQUEST, "JavaDoc [...]")
```

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/DocExtractor.kt
```diff
@@ -0,0 +1,56 @@
...
+            val entry = jarFile.getEntry("index.html")
+            check(!(entry == null || entry.isDirectory)) { "Invalid doc.jar!" }
+
+            val result = extractJavadoc(destination.toFile(), jarFile)
+            file.delete()
+            return result
+        }
+    }
+
+    private fun extractJavadoc(destination: File, jarFile: JarFile): Result<Unit, ErrorResponse> {
```
dzikoysk, line 35: Try to use `Path` instead of `File`, it's more consistent with the rest of sources and uses nio.

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/DocExtractor.kt
```diff
@@ -0,0 +1,56 @@
...
+        check(file.name.contains("doc.jar")) { "Invalid javadoc.jar! Name must contain: \"doc.jar\"" }
+
+        JarFile(file).use { jarFile ->
+            // Making sure we have an index.html file
+            val entry = jarFile.getEntry("index.html")
+            check(!(entry == null || entry.isDirectory)) { "Invalid doc.jar!" }
+
+            val result = extractJavadoc(destination.toFile(), jarFile)
+            file.delete()
+            return result
```
dzikoysk, line 31: Might be simplified with

```kotlin
return extractJavadoc(destination.toFile(), jarFile).also { /* delete */ }
```

Also, the `file` represents javadoc jar file, I don't think it should be deleted 🤔

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/DocExtractor.kt
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    private fun extractJavadoc(destination: File, jarFile: JarFile): Result<Unit, ErrorResponse> {
+        val entries = jarFile.entries()
+
+        try {
+            entries.asSequence().forEach { file ->
+                val javaFile = File(destination.absolutePath.toString() + File.separator + file.name)
```
dzikoysk, line 40: `destination.absolutePath.resolve(file.name)`

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/DocExtractor.kt
```diff
@@ -0,0 +1,56 @@
...
+            file.delete()
+            return result
+        }
+    }
+
+    private fun extractJavadoc(destination: File, jarFile: JarFile): Result<Unit, ErrorResponse> {
+        val entries = jarFile.entries()
+
+        try {
+            entries.asSequence().forEach { file ->
```
dzikoysk, line 39: `forEach` could be probably applied directly on `jarFile.entries()`?

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocFacade.kt
```diff
@@ -0,0 +1,226 @@
...
+class JavadocFacade internal constructor(
+    private val javadocFolder: Path,
+    private val mavenFacade: MavenFacade,
+    private val journalist: Journalist
+) : Journalist, Facade {
+
+    private val extractor = DocExtractor()
+
+    fun resolveRequest(repo: String, gav: String, extension: String, accessToken: AccessToken?): Result<JavadocResponse, ErrorResponse> {
+        val target = javadocFolder.resolve("${repo}${File.separator}${gav}")
```
dzikoysk, line 31: `javadocFolder.resolve(repo).resolve(gav)`

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocFacade.kt
```diff
@@ -0,0 +1,226 @@
...
+                ok(JavadocResponse(contentType, response))
+            } catch (e: NoSuchFileException) {
+                errorResponse(HttpCode.NOT_FOUND, "Resource not found!")
+            }
+        }
+
+        val newGav = constructGav(gav)
+        val lookUp = mavenFacade.findDetails(LookupRequest(accessToken, repo, newGav));
+
+        if (lookUp.isOk) {
```
dzikoysk, line 48: In general, we don't want to unwrap responses, use `lookup.map { details -> newValue }` instead

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocFacade.kt
```diff
@@ -0,0 +1,226 @@
...
+            val file = lookUp.get()
+            if (file.type === FileType.FILE) {
```
dzikoysk, line 50: You can also use `filter({ condition }, { error() })`` to filter `Result` and avoid multiple nested conditions

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocFacade.kt
```diff
@@ -0,0 +1,226 @@
...
+                if (!file.name.endsWith("-javadoc.jar")) {
+                    return  errorResponse(
```
dzikoysk, line 52: You can use `notFoundError(message)` utility function

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocFacade.kt
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    /**
+     * Constructs a new gav from the input gav, so we can handle paths that end with /index.html for example.
+     * This method can be extended further, to support even more ways to get the same javadoc with less URL parameters.
+     */
+    private fun constructGav(gav: String): String {
+        if (gav.endsWith("index.html")) {
+            val path = gav.substringBeforeLast("/index.html")
+            val split = path.split("/".toRegex())
```
dzikoysk, line 75: Split should probably use raw string instead of regex? 🤔

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocFacade.kt
```diff
@@ -0,0 +1,226 @@
...
+        val targetFolder = javadocFolder.resolve("${repo}${File.separator}${path}")
+        if (Files.exists(targetFolder)) {
+            return ok(Files.readAllLines(targetFolder.resolve("index.html")).joinToString(separator = "\n"))
+        }
+
+        Files.createDirectories(targetFolder)
+        val targetJar = targetFolder.resolve("doc.jar")
+
+        val findResult = mavenFacade.findFile(LookupRequest(token, repo, gav))
+        if (findResult.isOk) {
```
dzikoysk, line 108: As before, try to use `map`

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocResponse.kt
```diff
@@ -0,0 +1,13 @@
+package com.reposilite.plugin.javadoc
+
+
+class JavadocResponse(contentType: String, response: Any?) {
```
dzikoysk, line 4: Just

```kotlin
data class JavadocResponse(val contentType: String, val response: Any?)
```

dzikoysk: Also, what's the `response`, can we slightly precise this type or we really want to support everything here?

dzikoysk (changes requested):

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocFacade.kt
dzikoysk: As before, because currently there is a chance that you'll try to resolve `repo\g/a/v`, because `File.separator` is platform dependent

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocFacade.kt
```diff
@@ -0,0 +1,280 @@
...
+        </html>
+        """.trimIndent()
+
+        Files.writeString(index, source, Charsets.UTF_8)
+    }
+
+    override fun getLogger(): Logger =
+        journalist.logger
+
+    internal class DocExtractor {
```
dzikoysk, line 229: Just pull these methods to facade, without extractor class

reposilite-plugins/javadoc-plugin/src/main/kotlin/com/reposilite/plugin/javadoc/JavadocResponse.kt
```diff
@@ -0,0 +1,4 @@
+package com.reposilite.plugin.javadoc
+
+
+data class JavadocResponse(val contentType: String, val response: Any?) {}
```
dzikoysk, line 4: `{}` is redundant here

dzikoysk (changes requested):

dzikoysk:
You have to define shadow plugin:

* https://github.com/dzikoysk/reposilite/blob/8cb4f442c0d502b353ffdbd840e80df558084fde/reposilite-plugins/example-plugin/build.gradle.kts#L6

You can also merge main branch to adapt to latest changes. Then it should be fine, it could be a little bit simplified, but you're new in Kotlin, so there is no point to focus on that now. Good job :)

TOTHT0MI:
I think I'm done. Things that can be improved in the future:
- JavaDoc response may use a strict type instead of Any
- JavaDoc container index.html page may have a customizable title, download documentation button (jar, zip), etc.

(P.S: I'll probably implement the webhook plugin in the following days if this gets approved that is)

dzikoysk (approved):

dzikoysk:
Thanks for your contribution! ❤️

Issue #1246, opened 2022-04-22
LdapAuthenticator fails with "Could not identify one specific result after filter" 


shuawest:
### What happened?

When configuring reposilite against an LDAP server the connector fails to filter users when there is more than one user in the directory. 

Adding a second user in LdapAuthenicatorTest.kt demonstrates this error, resulting in a failed test with output:
```
LdapAuthenticatorTest > should authenticate existing ldap user() FAILED
    org.opentest4j.AssertionFailedError: Expected success, but result errored with: <ErrorResponse(status=400, message=Could not identify one specific result)> // <-- from second filter statement in LdapAuthenticator.kt
        at app//org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:39)
        at app//org.junit.jupiter.api.Assertions.fail(Assertions.java:134)
        at app//panda.std.ResultAssertions.assertOk(ResultAssertions.java:14)
        at app//com.reposilite.auth.LdapAuthenticatorTest.should authenticate existing ldap user(LdapAuthenticatorTest.kt:117)
```

The root cause is in the LdapAuthenticator.kt authenticate function.  The first directory search filters users based on the auth request name matched to the userAttribute. There is a second directory search after the user is authenticated where only the userFilter is applied, however the authentication request name is not used nor available to that filter. 
```
.flatMap { createContext(user = it.first, password = authenticationRequest.secret) } // try to authenticate user with matched domain namespace
.flatMap { it.search(userFilter, userAttribute) } // <-- this search results in all users being returned from ldap that match the 'Maven Users' filter, not the only the target user                 
.filter({ it.size == 1 }, { ErrorResponse(BAD_REQUEST, "Could not identify one specific result after filter") })
``` 

Adding the userFilter with userAttribute matched with the authentication request name, as done in the first search resolves the error and makes the ldap connector functional against a realistic directory. 
```
.flatMap { createContext(user = it.first, password = authenticationRequest.secret) } // try to authenticate user with matched domain namespace               
.flatMap { 
    it.search(
        "(&(objectClass=person)(${userAttribute}=${authenticationRequest.name})${userFilter})", // filter result with user-filter from configuration                
        userAttribute
    ) 
} 
.filter({ it.size == 1 }, { ErrorResponse(BAD_REQUEST, "Could not identify one specific result after filter") }) // only one search result allowed                
``` 

I will send a PR with the fix and updated test reproducer update.

### Reposilite version

3.x

[...]

PR #1247, opened 2022-04-22
GH-1246 Fix for ldap search filter

shuawest:
When configuring reposilite against an LDAP server the connector fails to filter users when there is more than one user in the directory. 

The root cause is in the LdapAuthenticator.kt authenticate function. The first directory search filters users based on the auth request name matched to the userAttribute. There is a second directory search after the user is authenticated where only the userFilter is applied, however the authentication request name is not used nor available to that filter.

Adding the userFilter with userAttribute matched with the authentication request name, as done in the first search resolves the error and makes the ldap connector functional against a realistic directory.


dzikoysk (approved):

dzikoysk:
Thanks for the PR ❤️ Looks fine, I'll merge this when builds pass all tests on CI

Issue #1287, opened 2022-05-08
Publish nightly versions of standalone jar files


dzikoysk:
### Request details

.

WithoutAName25:
My idea would be to build SNAPSHOT versions and publish them on [https://repo.panda-lang.org/#/snapshots](https://repo.panda-lang.org/#/snapshots). It should be possible to append `-SNAPSHOT` to the version in the `build.gradle.kts` if, for example, the environment variable `SNAPSHOT` is set to `true`.
The only problem with this approach is that if the current latest released version is, for example, `3.0.0` the next SNAPSHOT build would have the version `3.0.0-SNAPSHOT` but it should be `3.0.1-SNAPSHOT` if the next version to be released in the future is `3.0.1`.
One solution could be that we increase the version after each release instead of before each release. Another possible solution could be to always use just `3-SNAPSHOT` or `3.0-SNAPSHOT` or `3.x-SNAPSHOT` but I don't think that's how it's meant to be.

dzikoysk:
Maybe something like this could be helpful:

* https://github.com/researchgate/gradle-release

🤔

WithoutAName25:
I have tried the plugin in a test repo and it is basically designed to manage the complete versioning of a project. Do we want to automate the hole versioning stuff or do we only want to publish nightly standalone versions?

The relevant steps that the plugin does:
- remove `-SNAPSHOT` from the version
- commits the new version
- creates release tag with the current version for that commit
- increases the last number of the version (`1.0.0` -> `1.0.1` or `1.0.0-alpha.1` -> `1.0.0-alpha.2`) or set it to a predefined version
- commits the new version with the -SNAPSHOT flag

The problems with the plugin:
- researchgate/gradle-release#348
- only the version in `gradle.properties` will be updated
- some problems with kotlin gradle dsl, but it would work
- support for multi-project builds isn't complete

If we want to automate the versioning, I think it is better to update the version with github actions manually. Then we could update the version in all places.
If we only want to publish nighty standalone versions, we need another solution.

dzikoysk:
At first I thought that `nightly` release build as a standalone JAR could be enough, but after a while I think it might be just better option to provide standard snapshot builds. The reason why I think it might be just a good step for a project is that because this time we're also providing API for developers (plugins), so it might be quite useful to provide valid maven artifacts for them in case of any changes in the main codebase.

[...]

PR #1322, opened 2022-05-19
Gh-1287 Automatic versioning and snapshot builds

WithoutAName25:
The versioning is now based on git tags. To generate the next version, by default the last number will be increased. If you want to specify the next version explicitly there are next version markers (git tags with a suffix). By default the suffix is `-alpha` but because we are using alpha builds, I have changed it to `-next` (example marker: `3.0.0-next`). Another option would be `-snapshot` as suffix. These tags can be created manually or with `./gradlew markNextVersion -Prelease.version=2.0.0` / `./gradlew markNextVersion -Prelease.incrementer=incrementMinor`.

If you run the release workflow, all files containing the version are updated and commited with a tag for that version. Then the jar and the docker image is published.


dzikoysk:
Thanks!

I'd rename `release` workflow to `publish-releases` and `snapshot` to `publish-snapshots`, but despite of that, I think it's fine. I guess we just need to try to launch it and see what will happen 😳

WithoutAName25:
I have tested a few things but not everything, so I guess we have to try and hope that everything works as expected.

dzikoysk (approved):

Issue #1330, opened 2022-05-21
Snapshot builds don't have updated version in the internals


dzikoysk:
### What happened?

New release properly publish next version:

* https://maven.reposilite.com/#/snapshots/org/panda-lang/reposilite/3.0.0-alpha.26-SNAPSHOT

But the version in sources is not updated, would be great to have it fixed if possible. Probably this hook is not invoked for standard publish build:

https://github.com/dzikoysk/reposilite/blob/08c9a3fe3d85909bbb3727ca8dbf06d124538d69/build.gradle.kts#L41-L52

### Reposilite version

3.x

dzikoysk:
@WithoutAName25 could you take a look on this? :)

WithoutAName25:
These hooks are primarily to update for example the latest version in `README.md` on each full release (not snapshots). With this plugin there is no option to update some versions also for snapshots but I will try to find another solution for that.

WithoutAName25:
My first idea was to include the version in `META-INF/MANIFEST.MF` and then load it during runtime. But because the version is used in annotations, it has to be a compile time constant. [This](https://stackoverflow.com/questions/59692381/how-can-i-modify-java-source-code-in-my-gradle-project-prior-to-compilation-with/59692997#59692997) could work but it's maybe to complex for just getting the correct snapshot version at runtime.

dzikoysk:
So it won't work using axios plugin even if we'd create like a "fake" commit on CI with a next snapshot provided as the next release?

Speaking of the annotations, we can always replace the explicit version with a placeholder resolved during loading, so I think it wouldn't be a problem.

WithoutAName25:
It is not possible with axios plugin, even with "fake" commits. But if we don't need the version at compile time (no annotations with the version), I think the best approach is using `META-INF/MANIFEST.MF`

dzikoysk:
Ok, I think it's worth to give such approach a chance, I even prefer to have sources without explicit references to the current version.

WithoutAName25:
I think we need another solution because `VERSION` is used in some annotations were it's needed:
https://github.com/dzikoysk/reposilite/blob/08c9a3fe3d85909bbb3727ca8dbf06d124538d69/reposilite-backend/src/main/kotlin/com/reposilite/console/CommandExecutor.kt#L32-L33
[...]

PR #1344, opened 2022-05-24
GH-1330 Internal version for snapshot builds

WithoutAName25:
The contents of `src/template/kotlin` are copied to `src/generated/kotlin` and `@version@` gets replaced with the version.
I think `src/generated/kotlin` is better than `build/generated/kotlin` because otherwise the ide would show an error before the first compilation and after each `./gradlew clean`.

### DIFF
--- a/reposilite-backend/src/generated/kotlin/com/reposilite/ReposiliteProperties.kt
+++ b/reposilite-backend/src/generated/kotlin/com/reposilite/ReposiliteProperties.kt
@@ -1,0 +1,3 @@
+package com.reposilite
+
+const val VERSION = "3.0.0-alpha.26-SNAPSHOT"
--- a/reposilite-backend/build.gradle.kts
+++ b/reposilite-backend/build.gradle.kts
@@ -15,6 +15,7 @@
  */
 
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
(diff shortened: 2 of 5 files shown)


reposilite-backend/src/generated/kotlin/com/reposilite/ReposiliteProperties.kt
```diff
@@ -0,0 +1,3 @@
+package com.reposilite
+
+const val VERSION = "3.0.0-alpha.26-SNAPSHOT"
```
dzikoysk, line 3: If that's guaranteed to be generated during build, shouldn't the `generated` directory be excluded from git?

WithoutAName25: Yes, this could be excluded from git, the only disadvantage is that the ide shows an error after you clone the repository if you haven't build the project yet.

dzikoysk: I see, ok, let's keep it for now, maybe I'll add a note to this directory to describe why it's attached

dzikoysk (approved):

dzikoysk:
Thanks!

PR #1362, opened 2022-06-05
Fix wrong plugin description

mufinlive:
While browsing the reposilite website, I noticed that the plugin description for the Migration plugin was copied from the Swagger UI plugin. This is of course not intended, as they do completely different things. This PR aims to fix this issue.


dzikoysk (approved):

dzikoysk:
Thanks!

Issue #1397, opened 2022-06-17
Unable to setup reverse proxy through nginx


anibmurthy:
### What happened?

Hello, I am trying to setup reverse proxy using nginx through docker-compose and getting this error:
```
nginx         | 2022/06/17 06:06:06 [error] 24#24: *1 connect() failed (111: Connection refused) while connecting to upstream, client: 192.168.0.1, server: , request: "GET / HTTP/1.1", upstream: "http://127.0.0.1:9090/", host: "localhost"
```
Here is the `docker-compose` file:
```
version: '3'

services:
  nginx:
    image: nginx:latest
    container_name: nginx
    volumes:
    - ./nginx.conf:/etc/nginx/nginx.conf
    ports:
    - 80:80

  reposilite:
    image: dzikoysk/reposilite:3.0.0-alpha.15
    container_name: local-repo
    environment:
    - REPOSILITE_OPTS=--token=admin:admin --port=9090
    ports:
    - 9090:9090
    restart: always
```
And nginx configuration:
```
user www-data;
worker_processes auto;
pid /run/nginx.pid;
include /etc/nginx/modules-enabled/*.conf;

events {
   worker_connections 1024;
}

http {
   server {
      listen 80;
      server_name localhost 127.0.0.1;

      location / {
         proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header Host $host;
         proxy_set_header X-Real-IP  $remote_addr;
         proxy_pass              http://localhost:9090/;
      }
   }
}
```
With above setup, direct link http://localhost:9090/ loads reposilite homepage as expected. However, does not work with http://localhost/.

P.S.: Thanks for this wonderful app. It perfectly suits my requirement.

### Reposilite version

3.x

### Relevant log output

```shell
nginx         | /docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
nginx         | /docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
nginx         | /docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
nginx         | 10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
nginx         | 10-listen-on-ipv6-by-default.sh: info: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
[...]

PR #1411, opened 2022-06-21
GH-1397 Explain the IP to use for Nginx reverse proxy

nik-x:
Added some notes on what IP address to use when using Nginx as reverse proxy in different scenarios as we previously had some people getting confused about it.


dzikoysk (approved):

dzikoysk:
Thanks!

Issue #1423, opened 2022-06-26
Add snapshot support to javadoc plugin


joshuasing:
### Request details

It would be very helpful if the javadoc plugin supported snapshots.

Currently going to `https://repo.domain.com/javadoc/snapshots/com/domain/artifact/0.1.0-SNAPSHOT/` will result in
```json
{"status":404,"message":"Cannot find com/domain/artifact/0.1.0-SNAPSHOT/artifact-0.1.0-SNAPSHOT-javadoc.jar in local and remote repositories"}
```
because the `.jar` file it is trying to find is formatted as `artifact-0.1.0-20220626.075216-1-javadoc.jar` instead of `artifact-0.1.0-SNAPSHOT-javadoc.jar`

If I have the time over the following days and this is not already being worked on, I may be able to create a pull request to add it.
Thank you!

PR #1441, opened 2022-06-30
GH-1423 Add snapshot support to javadoc-plugin

joshuasing:
Adds the ability to use the javadoc-plugin with snapshot versions.
See #1423 for more information

I am not very familiar with the code base so if there is a better way of doing this, please let me know so I can improve.


dzikoysk:
Thanks!

PR #1459, opened 2022-07-06
GH-1459 Fix CardMenu dropdown selection 

joshuasing:
Fixes the CardMenu dropdown selection, previously it tried to use a method that didn't exist and would just throw errors.


dzikoysk (approved):

dzikoysk:
Thanks!

PR #1467, opened 2022-07-08
GH-1467 Make the plugin directory configurable

Saladoc:
Since plugins are technically a part of the code, not the configuration, it should be possible to separate them from the workdir.


dzikoysk (approved):

dzikoysk:
Thanks!

PR #1546, opened 2022-09-10
Update FrontendFacade.kt

Michele0303:
xss vulnerability fixed

### DIFF
--- a/reposilite-backend/src/main/kotlin/com/reposilite/frontend/FrontendFacade.kt
+++ b/reposilite-backend/src/main/kotlin/com/reposilite/frontend/FrontendFacade.kt
@@ -29,7 +29,7 @@ class FrontendFacade internal constructor(
 
     private val resources = HashMap<String, String>(0)
     private val uriFormatter = Regex("/+") // exclude common typos from URI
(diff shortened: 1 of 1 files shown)


dzikoysk:
`createNotFoundPage` is internal method that is used twice in our code in such scenarios:

obraz

Once as empty string, and second one to display `message` from `ErrorResponse` data object that is also internal wrapper for `status code + http message`. I don't think we have any XSS vulnerability here 🤔

Michele0303:
Hi, I found the vulnerability in my friend's site.

https://pastebin.com/7FfT2Fb9

Il giorno sab 10 set 2022 alle ore 23:28 dzikoysk ***@***.***>
ha scritto:

Michele0303:
I sent you an email with PoC

dzikoysk:
I see, nice catch, indeed you're right! I think we should also support characters like `.`, `-`, `~`, because those are common in gav names/versioning.

dzikoysk (approved):

dzikoysk:
Thanks! It's not that urgent as it'd require targeted attack on given repository & user, so I'll release a new version tomorrow, because it's quite late + I want to fix something else in 3.0.3

Issue #1604, opened 2022-10-14
Mute subscriber that tries to write to closed output


dzikoysk:
### Request details

![obraz](https://user-images.githubusercontent.com/4235722/195849061-491c5202-0208-4586-be52-b8155d6b7b05.png)

Probably caused by: https://github.com/dzikoysk/reposilite/blob/main/reposilite-backend/src/main/kotlin/com/reposilite/console/infrastructure/ConsoleWebSocketHandler.kt#L69

Anuran12:
@dzikoysk I would like to do this.. but don't know how to solve this... please guide

dzikoysk:
I think that simple try-catch around ctx.send() for ClosedChannelException could be enough, this error is pretty random and it's hard to reproduce it, we will just observe the result :)

Anuran12:
> I think that simple try-catch around ctx.send() for ClosedChannelException could be enough, this error is pretty random and it's hard to reproduce it, so we will just observe the result :)

Maybe I could not do it.. you better to assign someone else.. sorry to bother 😢

P-DR0ZD:
Can I be assigned I think I could solve it

P-DR0ZD:
What should be done when the exception is caught?

dzikoysk:
I think we can safely mute it by just logging it at `DEBUG` level? 🤔

PR #1607, opened 2022-10-14
GH-1604 Mute subscriber that tries to write to closed output

P-DR0ZD:
I caught the error and added the debug logging as asked for in issue #1604. please check if the logging was done correctly

### DIFF
--- a/reposilite-backend/src/main/kotlin/com/reposilite/console/infrastructure/ConsoleWebSocketHandler.kt
+++ b/reposilite-backend/src/main/kotlin/com/reposilite/console/infrastructure/ConsoleWebSocketHandler.kt
@@ -67,11 +67,10 @@ internal class CliEndpoint(
                             journalist.logger.info("CLI | $identifier accessed remote console")
 
                             val subscriberId = journalist.subscribe {
-                                try{
+                                try {
                                     ctx.send(it.value)
-                                }
-                                catch(e: ClosedChannelException) {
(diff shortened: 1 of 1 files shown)


reposilite-backend/src/main/kotlin/com/reposilite/console/infrastructure/ConsoleWebSocketHandler.kt
```diff
@@ -66,7 +66,12 @@ internal class CliEndpoint(
                             journalist.logger.info("CLI | $identifier accessed remote console")
 
                             val subscriberId = journalist.subscribe {
-                                ctx.send(it.value)
+                                try{
+                                    ctx.send(it.value)
+                                }
+                                catch(e: java.nio.channels.ClosedChannelException) {
```
dzikoysk, line 72: Could you import this class? :)

dzikoysk (changes requested):

reposilite-backend/src/main/kotlin/com/reposilite/console/infrastructure/ConsoleWebSocketHandler.kt
P-DR0ZD: No Problem I just imported it.

dzikoysk (approved):

dzikoysk:
Thanks!

Issue #1654, opened 2022-12-11
Enhance management panel with dashboard


dzikoysk:
### Request details

.

PR #1655, opened 2022-12-11
GH-1654 Implement dashboard

dzikoysk:
obraz

SJ PW 2022


ThraaxSession:
Hi @dzikoysk wouldn't it be more accessible of the padding between resolved requests is smaller? It takes a lot of space.

dzikoysk:
I'm going to add one more chart, so I'll probably try to reduce  its height 🤔

dzikoysk:
@ThraaxSession 

obraz

What do you think?

dzikoysk:
Nice :)

dzikoysk:
Thanks 💜 Yeah, I think it might be a good idea to release a next version with those charts.

PR #1667, opened 2022-12-28
Set default S3 region to prevent crashes

### LINKED ISSUE #1666: Crash when using third-party S3 storage provider and not providing a region
### What happened?

I've tried setting up reposilite using Backblaze as my storage provider, and have left the region field empty as it was apparently optional and it is already part of the endpoint override URL, this has however led to reposilite crashing in a non-recoverable form.

From experience creating my own S3 implementation recently this is because the region is part of the signed authorization header S3 uses (see [the documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html)), so it must always be set to some dummy value for third-party storage providers. This can be tested by for example setting the region to `reposilite`.

### Reposilite version

3.x

### Relevant log output

```shell
https://files.lostluma.net/WHCvJq.txt
```

LostLuma:
Fixes #1666 by always setting a region on the S3 client.

I've not been able to test this locally (even with the docker-compose config), and have never written Kotlin, so I am unsure whether this works. If another solution is preferred feel free to close this PR.

### DIFF
--- a/reposilite-backend/src/main/kotlin/com/reposilite/storage/s3/S3StorageProviderFactory.kt
+++ b/reposilite-backend/src/main/kotlin/com/reposilite/storage/s3/S3StorageProviderFactory.kt
@@ -27,6 +27,8 @@ class S3StorageProviderFactory : StorageProviderFactory<S3StorageProvider, S3Sto
 
         if (settings.region.isNotEmpty()) {
             client.region(Region.of(settings.region))
+        } else {
+            client.region(Region.of("reposilite"))
         }
 
         if (settings.endpoint.isNotEmpty()) {


dzikoysk (approved):
Thanks! Looks fine for me, as a tip I can say you can use `ifEmpty { altText }` in Kotlin, so e.g.:

```kotlin
if (settings.region.isNotEmpty()) {
    client.region(Region.of(settings.region))
} else {
    client.region(Region.of("reposilite"))
}
```

could look like this:

```kotlin
client.region(Region.of(settings.region.ifEmpty { "reposilite" }))
```

Anyway, I'm fine with current impl so I'm going to merge this :)

PR #1694, opened 2023-01-17
GH-1694 Create Caddy guide

JorenSix:
I use Caddy as a reverse proxy to reposilite. The amount of configuration needed to run a reverse proxy over HTTPS is spectacularly lower than nginx. This might be of interest to others as well and might fit in many cases.

### DIFF
--- a/reposilite-site/data/guides/integrations/caddy.md
+++ b/reposilite-site/data/guides/integrations/caddy.md
@@ -1,0 +1,58 @@
+---
+id: caddy
+title: Caddy
+---
+
+Caddy is a http server with support for reverse proxy and automatic https. The automatic https feature 
+takes care of requesting, installing and updating SSL certificates which means that you need much less configuration settings
+or maintenance compared with e.g. Nginx.
+
+Note that websockets are also transparent in the case of reverse proxying.
+
+First [download and install caddy](https://caddyserver.com/docs/install) and modify your caddy configuration file, 
+typically found in `/etc/caddy/Caddyfile`.
+
+```json5
+
+(revproxy) {
+         #compress responses
+         encode zstd gzip
+
+         #redirect from http to https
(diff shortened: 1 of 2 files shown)


reposilite-site/data/guides/integrations/caddy.md
```diff
@@ -0,0 +1,58 @@
...
+        import revproxy 8080
+}
+```
+
+You might need to modify a few things:
+
+* If Reposilite is running on an other machine, replace `localhost` in the `reverse_proxy localhost:{args.0}` statement to the IP or host of that machine.
+* Here the `domain.com` refers to the domain name in used. Typically this needs to be changed.
+* The port of the Reposilite instance here is 8080, this needs to match your configuration.
+* With URL-rewriting it is also possible to redirect a directory `/reposilite' to your instance. I leave this as an excercise for the reader.
```
dzikoysk, line 42: I'd replace `I leave this as an excercise for the reader.` with a link to Caddy's guide about that 😅

dzikoysk (changes requested):
Thanks for your contribution! We have some Caddy users, so indeed it might be good idea to provide official guide. I'll merge that after a hot-fix to this custom base path docs reference :)

dzikoysk:
Oh, one more thing, you should also link it in `guides.js`:

https://github.com/dzikoysk/reposilite/blob/2139b473d22a242e5bbd38f4b0c05ed3835fe10f/reposilite-site/data/guides/guides.js#L55-L61

JorenSix:
I have changed the guides.js file and the caddy.md file to reflect the requested changes. There is now a link to the caddy documentation page and the page is linked.

dzikoysk (approved):

dzikoysk:
Thanks!

PR #1726, opened 2023-02-07
GH-1726 Bugfix for version comparator + tests

bergerkiller:
I ran into a problem with javadoc /latest selecting version 1.19-SNAPSHOT over 1.19.3-SNAPSHOT, and after a bit of digging the problem was found to be a simple mistake. I've added a couple unit tests so I could verify that my fix had the intended effect.

One test case was modified to suit this new ordering, I don't know if that is a problem or not. By the rules of the algorithm 1.1 and 1.1.2 (release) should be newer than 1.1-SNAPSHOT, so this causes all words to be sorted to the front instead of the bottom. (might need a re-build as the auto test failed earlier because of it)


dzikoysk (approved):

dzikoysk:
Thanks! I really like the test following our test-style. Also, sorry for late response, I kinda lost the notification about this PR 👀

PR #1734, opened 2023-02-26
GH-1734 Add a healthcheck to the Dockerfile

rtm516:


dzikoysk:
Thanks for PR :) I see 2 issues with those changes:

1. Port in Reposilite is configurable, so we cannot hard code it to 8080 :/
2. I'd start after 15s, because budget ARM devices might be too slow to launch whole instance in 5s

rtm516:
> Thanks for PR :) I see 2 issues with those changes:
> 
> 1. Port in Reposilite is configurable, so we cannot hard code it to 8080 :/
Not sure how to handle this as there are multiple ways of configuring the port, the configuration file can likely be read and port pulled that way but the command line argument is more complex.
Needs more thinking to handle that. Although most docker users probably wont be changing the port since you can redirect it using docker.

> 2. I'd start after 15s, because budget ARM devices might be too slow to launch whole instance in 5s
Ajusted

dzikoysk:
I guess we could generate at startup some kind of a temp file with address pointing to the local instance of Reposilite 🤔

rtm516:
Sounds like a good idea, I might not get chance to do that until the weekend, feel free to edit this PR if you want to have a go. I can make the health check changes if you make just the address file dump.

dzikoysk:
Sorry, I forgot that there was an enhancement request in this PR. The address is available in `.local/reposilite.address` file now.

dzikoysk (approved):

dzikoysk:
Thanks, let's see how it works :)

Issue #1758, opened 2023-03-24
Repository Name for Gradle


williambl:
### Request details

On the right-hand-side there is a card, which shows snippets to add the repository to Maven/Gradle/SBT. For Maven, this gives the repository a human-readable name:
![<repository> <id>reposilite-repository-releases</id> <name>Reposilite Repository</name> <url>http://localhost:8081/releases</url> </repository>](https://user-images.githubusercontent.com/5115825/227642574-5800eccc-20a3-4ee5-b595-2b3f6ac00837.png)

But it doesn't for either Gradle language:
![maven { url = uri("http://localhost:8081/releases") }](https://user-images.githubusercontent.com/5115825/227642641-fa8ea03d-8a16-4078-bb63-42267e91d1b8.png)
![maven { url "https://maven.willbl.dev/silver" } ](https://user-images.githubusercontent.com/5115825/227642653-73719703-a56a-4439-8a6e-292d9a714751.png)

Even though Gradle does have support for human-readable names for repositories.

It'd be nice if Reposilite showed human-readable names for Gradle snippets too.

This is implemented in #1759.

PR #1759, opened 2023-03-24
GH-1758 Add human-readable names for Gradle snippets

williambl:
See #1758. A simple change.

Before:
Screenshot 2023-03-24 at 21-00-51 {{REPOSILITE TITLE}}
Screenshot 2023-03-24 at 21-00-56 {{REPOSILITE TITLE}}

After:
Screenshot 2023-03-24 at 21-02-29 {{REPOSILITE TITLE}}
Screenshot 2023-03-24 at 21-04-04 {{REPOSILITE TITLE}}

### DIFF
--- a/reposilite-frontend/src/components/card/RepositorySnippet.vue
+++ b/reposilite-frontend/src/components/card/RepositorySnippet.vue
@@ -41,11 +41,13 @@ defineProps({
 </pre>
 <pre v-else-if="configuration.lang === 'groovy'">
 maven <CodeBrackets start="{" end="}">
+    name <CodeString>{{ data.title }}</CodeString>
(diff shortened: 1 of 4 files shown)


dzikoysk:
Thanks for PR! Yeah, it makes sense, but I'm wondering if that's the name we should use. In Gradle, given name is also kinda like an id, so it'd be nice to add the repo name there as well. Also, Gradle uses this strange format `myRepositoryName` - in tasks, but also in the guide, e.g.:

* https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:handling_credentials

So maybe we should define it like that 🤔

dzikoysk (approved):

dzikoysk:
Thanks :)

PR #1760, opened 2023-03-29
NixOS installation guide

rebarbora-mckvak:
I am by no means expert on NixOS, but I created these scripts to install Reposilite on NixOS.


dzikoysk (approved):

dzikoysk:
I have no idea how NixOS works and how to use it, but because you've successfully launched it with this guide, I think I can approve that. In case of any issues, more experienced NixOS users may try to improve that if needed :) Thanks!

PR #1780, opened 2023-04-05
Add SVG as supported javadoc type

JFronny:
Some javadoc doclets and taglets produce SVG images. This PR adds support for them.
Tested by publishing a project using one such doclet to an instance of reposilite with this change and verifying that it is displayed correctly.
There is one oddity in the embedded view: links in the SVG do not work, while links outside it do.
This is, however, not an issue in the raw view.


dzikoysk (approved):

dzikoysk:
Thanks!

PR #1817, opened 2023-05-07
Fix badges having a black background if no custom color is provided

Sytm:
If no custom color is provided the returned SVG document looks like this:
```xml
<g clip-path="url(#r)">
    <rect width="98" height="20" fill="#555"/>
    <rect x="98" width="49" height="20" fill="#null"/>
    <rect width="147" height="20" fill="url(#s)"/>
</g> 
```
with an fill color of `#null` which makes the background black
Example:


dzikoysk (approved):

dzikoysk:
Thanks!

PR #1851, opened 2023-06-14
Prettier uptime format

OliverSchlueter:
Added days to the uptime in the frontend dashboard.
Made the uptime in the status command the same format (`2d 8h 42min 5s`).


dzikoysk:
Thanks!

PR #1907, opened 2023-08-26
Disable auto complete for consoleInput

Fallen-Breath:
These autocomplete pop-ups are very annoying. They can prevent you from using the `↑` or `↓` keys to switch within the command history.

image

In my opinion, the command input box is not really a form. It's an interactive command input, which doesn't need any browser-provided autocomplete.

So, here's a quick fix for that: set `autocomplete="off"` and that's it.


dzikoysk (approved):

dzikoysk:
LGTM, thanks! :)

PR #1926, opened 2023-09-06
GH-1926 Fix tinylog file size policy

dkramer95:
The `writerFile.size` has no effect as I've observed the logs far exceeding the default `100MB` limit.

This changes to use the [documented](https://tinylog.org/v2/configuration/#policies) `size` policy.


dzikoysk (approved):

dzikoysk:
Thanks for PR! I've also lately noticed this issue on my personal instance, but I forgot to report this. :skull:

PR #1934, opened 2023-09-18
GH-1934 Check if user/group already exist in entrypoint.sh

houkunlin:
Fix the error when starting the container for the second time： addgroup: The group `reposilite' already exists.

image
image


dzikoysk (approved):

dzikoysk:
Thanks!

PR #1936, opened 2023-09-18
Update Dockerfile Add EXPOSE 8080

houkunlin:
Fixed the problem of missing public port configuration, causing Docker UI to be unable to recognize the application service port

image


dzikoysk:
Hmm... I think we had it in the past, but then we've added support for a custom port in the container. I guess it's fine to restore it as most of the people use the default one. Thanks :)

PR #1938, opened 2023-09-18
Fix reposilite not starting using docker

SageSphinx63920:
The current nightly isn't start, cause of some bash script problems. This pr fixes them


dzikoysk (approved):

dzikoysk:
Thanks!

PR #1982, opened 2023-11-29
GH-1982 Make `organizationWebsite` open in new tab

UpcraftLP:
This makes it so that the organization website link opens in a new tab when clicked, rather than replacing the current tab.


dzikoysk:
Thanks!

PR #2018, opened 2024-01-06
Add "no-refetch-timeout" for StoragePolicy.PRIORITIZE_UPSTREAM_METADATA

mkrueger92:
This commit adds an option to define a timeout until no re-fetch of maven-metadata.xml is done.

This gives the possibility to use the local copy until the defined period is over. Especially on connections with "huge" delays this setting improves build times and still has a deprecation mechanism for maven-metadata.xml content.

The default value is 0 seconds (it fetches the file every time).


dzikoysk:
Thanks! It makes sense, so I'd like to merge this today ([3.5.3](https://github.com/dzikoysk/reposilite/issues/2022)). I refactored it a bit on my own to save some time for both of us :)

PR #2042, opened 2024-02-01
GH-2042 Fix logic in isMetadataFileValid

mkrueger92:
The logic was in reverse. The file is valid when it is older than the defined time-range.

This commit reverses the logic and makes the feature usable.


dzikoysk:
You're right, my bad & thanks :)

dzikoysk (approved):

Issue #2054, opened 2024-02-14
401 Invalid Authorization Credentials using LDAP auth


mstup:
### What happened?

I tried Reposilite 3.5.6 intergrate with OpenLDAP 2.4.57
On user login Reposilite reports error - "401 Invalid Authorization Credentials".

OpenLDAP log on Reposilite login:
```
ldap_1                | 65ccb304 conn=1021 fd=12 ACCEPT from IP=172.18.0.5:43118 (IP=0.0.0.0:389)
ldap_1                | 65ccb304 conn=1021 op=0 BIND dn="cn=readonly,dc=domain,dc=com" method=128
ldap_1                | 65ccb304 conn=1021 op=0 BIND dn="cn=readonly,dc=domain,dc=com" mech=SIMPLE ssf=0
ldap_1                | 65ccb304 conn=1021 op=0 RESULT tag=97 err=0 text=
ldap_1                | 65ccb304 conn=1021 op=1 SRCH base="ou=people,dc=domain,dc=com" scope=2 deref=3 filter="(&(objectClass=posixAccount)(uid=mykola))"
ldap_1                | 65ccb304 conn=1021 op=1 SRCH attr=uid
ldap_1                | 65ccb304 conn=1021 op=1 SEARCH RESULT tag=101 err=0 nentries=1 text=
ldap_1                | 65ccb304 conn=1022 fd=13 ACCEPT from IP=172.18.0.5:43120 (IP=0.0.0.0:389)
ldap_1                | 65ccb304 conn=1022 op=0 BIND dn="uid=mykola,ou=people,dc=domain,dc=com" method=128
ldap_1                | 65ccb304 conn=1022 op=0 BIND dn="uid=mykola,ou=people,dc=domain,dc=com" mech=SIMPLE ssf=0
ldap_1                | 65ccb304 conn=1022 op=0 RESULT tag=97 err=0 text=
ldap_1                | 65ccb304 conn=1022 op=1 SRCH base="ou=people,dc=domain,dc=com" scope=2 deref=3 filter="(&(objectClass=posixAccount)(uid=mykola)(memberOf=cn=users,ou=reposilite,dc=domain,dc=com))"
ldap_1                | 65ccb304 conn=1022 op=1 SRCH attr=uid
ldap_1                | 65ccb304 conn=1022 op=1 SEARCH RESULT tag=101 err=32 nentries=0 text=
```
We have two binds: 
1. conn=1021 
Works fine for connect **readonly** user
[...]

Issue #2005, opened 2023-12-31
Reposilite 4.x wishlist 


dzikoysk:
### About
Reposilite 3.x is a stable piece of software and for most of the time it serves you well. Of course, some decisions I made during 2.x -> 3.x rewrite were good, and some of them are a bit of a problem these days. I think that's a good time to start summing up the 3.x release line, so we can prepare for the next major version. Fortunately, this time we're more likely talking about an evolution, not a revolution - the backend module is much easier to maintain, Kotlin really paid off here. Sadly, I can't say it about frontend in JS + Vue 3. 

### Roadmap

Some high level estimation of a roadmap for Reposilite 4.x 💕 

| Date | Description |
| :-: | :-: |
| Q1 & Q2 2024 | Get some feedback from the community about current state of the project and possible enhancements in the 4.x |
| Q3 2024 | Refine ideas, prepare project |
| Q4 2024 | Development phase |
| Q1 2025 | Release 4.x 🎉 |

### General goals

| Goal | Result |
| :-: | :-: |
| Include long overdue improvements that couldn't be addressed in 3.x | Improve the offer |
| Remove tech debt and reduce the complexity of frontend implementation | Easier to maintain |
| Prioritize simplicty and lightweightness | Keep the core advantage over alternatives |

### RFCs

| Priority | Domain | Description | Status |
| :---: | :---: | :--- | :---: |
| P1 | Tokens → Users | Reposilite should introduce user system. Users should be able to generate personal tokens on their own. | ✔️ |
| P1 | Authentication | Thanks to the new user system in Reposilite, we should be able to support the following auth methods in the UI: <br/><ul><li><a href="https://github.com/dzikoysk/reposilite/issues/1011#issuecomment-1016497861">OAuth, OIDC</a></li></ul> | 📢 |
| P1 | UI / Dashboard | Maintaining current REST API + standalone frontend module requires a lot of work. With [JTE](https://jte.gg/)-like template engine + some lightweight library like HTMX + Tailwind, we could probably do the same thing with a better result (and improve mobile UX)  | ❔ |
| P2 | Security | In-app alerting system to notify users in case of diagnosed security related issues (like CVEs) | ✔️ |
| P2 | Licensing | Reconsider license change to protect project from SaaS-like actors | ❔ |
| P2 | General | Simplify sources: Let's try to simplify & reduce sources having some more experience | ✔️ |
| P2 | Build-systems | Prepare a better support on API level for other build systems / protocols | ❔ |
[...]

PR #2055, opened 2024-02-15
GH-2054 Cover #2054 with test case

dzikoysk:


reposilite-backend/src/test/kotlin/com/reposilite/auth/LdapAuthenticatorTest.kt
```diff
@@ -18,123 +18,191 @@ package com.reposilite.auth
...
-                name = token.name,
-                secret = secret
+            ldapServer.add("dn: cn=users,dc=domain,dc=com", "ou: people", "objectClass: posixAccount", "memberOf: ou=people,dc=domain,dc=com")
+            ldapServer.add("dn: cn=readonly,dc=domain,dc=com", "ou: reposilite", "uid: readonly", "objectClass: posixAccount", "memberOf: cn=users,ou=reposilite,dc=domain,dc=com")
+
+            ldapServer.add("dn: uid=mykola,ou=people,dc=domain,dc=com", "ou: reposilite", "objectClass: posixAccount", "memberOf: cn=users,ou=reposilite,dc=domain,dc=com")
+        }
+
+        @Test
+        fun `should authenticate mykola`() {
```
mstup, line 180: @dzikoysk Please correct me if I'm wrong but understand this test like only "mykola" is allowed to login. But as I described it in the ticket, only "readonly" should be allowed to connect LDAP and all other should be blocked. This is how BindDN works.

dzikoysk: Okay, so that's not how I understood the issue. From the logs you've provided I thought you can successfully login with readonly (conn=1021), but you cannot do this with mykola (conn=1022). Could you specify your LDAP server config that I can use in this test to go through your issue? I feel like it can be done in multiple ways and I'm currently trying to guess what setup you actually have :thinking:

mstup: [ldap.txt](https://github.com/dzikoysk/reposilite/files/14311320/ldap.txt)

mstup: [ldap.txt](https://github.com/dzikoysk/reposilite/files/14311332/ldap.txt)

mstup: > Okay, so that's not how I understood the issue. From the logs you've provided I thought you can successfully login with readonly (conn=1021), but you cannot do this with mykola (conn=1022).

Yes, you understand it correct. User "mykola" shouldn't have expected to LDAP server. Access is granted only for user specified in Reposilite configuration and none other.

dzikoysk:
So this is your configuration:

```
"baseDn": "ou=people,dc=domain,dc=com",
"searchUserDn": "cn=readonly,dc=domain,dc=com",
"searchUserPassword": "nopassword",
"typeAttribute": "posixAccount",
"userAttribute": "uid",
"userFilter": "(memberOf=cn=users,ou=reposilite,dc=domain,dc=com)",
"userType": "PERSISTENT"
```

Let's go through this:
1. Base distinguished name has been set to `ou=people,dc=domain,dc=com`, it means that search base will only examine `ou=people` subtree in the `dc=domain,dc=com` directory
2. Search user has been set to `cn=readonly,dc=domain,dc=com` - this user will be used by Reposilite to check credentials of users that are trying to login to Reposilite
3. Password of search user (`cn=readonly,dc=domain,dc=com` in this case)
4. `typeAttribute` objectClass filter, so we will only check for records that `posixAccount` objects
5. `userAttribute` property that will contain username for a token that Reposilite will generate, e.g. `uid=dzikoysk`
6. User filter selects entries that are allowed to use Reposilite, in this case it has to contain `(memberOf=cn=users,ou=reposilite,dc=domain,dc=com)` trait

Having that in mind, I've updated our test:
1. `cn=readonly,dc=domain,dc=com` is properly authenticated on Reposilite's side and can perform search queries
2. `uid=mykola,ou=people,dc=domain,dc=com` is not authenticated, because `memberOf` is not matched
3. `uid=panda,ou=people,dc=domain,dc=com` is authenticated, because it's a part of Reposilite group

mstup:
Yes that's correct now. 
No one other user shouldn't try to bind LDAP server only one defined in Reposilite settings. Users authorization is made by search user is **memberOf** certain OU. If user is memeber of certain OU the next step is password validation.

This code which you have now (v 3.5.6) is suitable for the case when each user connect server separately. But on user interface there is way to setup BindDN connection (only one user have to do all work).

Here is good screenshots for Gitea to understand what I'm talking about.

==============>  Simple Auth
image



==============>  BindDN
image

dzikoysk:
Okay, I think I got what you assume the bug is here. You're trying to reuse `cn=readonly,dc=domain,dc=com` as an actual user, not the authorization credentials for server-side ldap connection that queries your directories, right? 

If that's correct, then you should modify your `userAttribute` in Reposilite's configuration - probably to `cn`, because that was the expected name you used here: https://github.com/dzikoysk/reposilite/issues/2054. There's no option to login to the dashboard without the username - that's because Reposilite does not really have a concept of users, it's directly mapped to access token on our side & the name is automatically fetched from the LDAP settings (we don't have complete registration form, or something like that, so we can't simply skip this attribute for now).

Also, if you'll confirm this is valid, I see that we could improve our integration - especially considering changes that are coming in 4.x:

* https://github.com/dzikoysk/reposilite/issues/2005

mstup:
> Okay, I think I got what you assume the bug is here. You're trying to reuse `cn=readonly,dc=domain,dc=com` as an actual user, not the authorization credentials for server-side ldap connection that queries your directories, right?

Yes, this is what page design is propose to do (at least how I understood that with my previous experience with Gitea page design of integration with LDAP).

mstup:
> the name is automatically fetched from the LDAP settings

But, user which connect to server again to fetch the name should remains the first one, but not this one the name of which you want to receive.

1. You connect first time with **readonoly** user and checking if user **mykola** exist.
2. If user mykola exist you connected second time to check his memberOf value. But to do this needs again connect to LDAP as **readonly** user. Any other user can't login to LDAP server it is prohibited by LDAP server configuration.

mstup:
@dzikoysk 
As I understood this code and as I already mentioned in the ticket. I see problem in this **[line](https://github.com/dzikoysk/reposilite/blob/4f5852515018d571ead657eb421a0b22b64dbff4/reposilite-backend/src/main/kotlin/com/reposilite/auth/LdapAuthenticator.kt#L89C1-L89C38)**.
Method **[createDirContext](https://github.com/dzikoysk/reposilite/blob/4f5852515018d571ead657eb421a0b22b64dbff4/reposilite-backend/src/main/kotlin/com/reposilite/auth/LdapAuthenticator.kt#L136)** preparing connection parameters and execute method [InitialDirContext](https://github.com/dzikoysk/reposilite/blob/4f5852515018d571ead657eb421a0b22b64dbff4/reposilite-backend/src/main/kotlin/com/reposilite/auth/LdapAuthenticator.kt#L145) and this is the second attempt to connect to LDAP server with user's received credentials (mykola's credentials).
Why?

dzikoysk:
Like I said before, this is the limitation of our current implementation in 3.x - password is required to create a new secret associated with generated access token. What you're asking for is a passwordless logging form for users working in the internal network - we don't have a concept of users nor the token management panel in the UI.

mstup:
Probably the context creation operation is also doing connect to LDAP server.

mstup:
Yes, looks like acceptable solution at this moment.

mstup:
Tested and works fine. Thank you.

PR #2100, opened 2024-04-18
Confusing token with secret.

iSilviu-debug:


dzikoysk:
This is the naming convention we use across all docs and internal implementation, so token means token's name and secret is the value of the token. It's a bit confusing, especially when using with user-based forms like the one in this example. It'll be improved in 4.x with a new user system, but for now we have to keep it like that.

PR #2148, opened 2024-06-26
GH-2147 Fix for manager permission

### LINKED ISSUE #2147: Fix permission in fake api
### What happened?

When signing in using fake api, user only gets read permission instead of manager

### Reposilite version

3.x

### Relevant log output

_No response_

Astahiell:
Resolves #2147


dzikoysk:
Thanks, I missed that :heart_hands:

PR #2171, opened 2024-07-11
GH-2171 Priority based authenticators

xanderume:


reposilite-backend/src/main/kotlin/com/reposilite/auth/AuthenticationFacade.kt
```diff
@@ -46,6 +46,7 @@ class AuthenticationFacade(
 
     fun registerAuthenticator(authenticator: Authenticator) {
         this.authenticators.add(authenticator)
+        this.authenticators.sortWith(compareByDescending { it.priority() })
```
dzikoysk, line 49: I think we should move to some sort of a sorted collection, e.g. sorted set, because in this case sorting is only applied when we're adding a new authenticator via this method. It's a bit inconsistent considering that we're also passing some of the authenticators via constructor.

reposilite-backend/src/main/kotlin/com/reposilite/auth/Authenticator.kt
```diff
@@ -27,6 +27,8 @@ interface Authenticator {
 
     fun enabled(): Boolean
 
+    fun priority(): Int = 0
```
dzikoysk, line 30: I think it's better to have a priority as a floating point number - we don't have a use-case for it yet, but it's a bit more flexible by design, because it might be helpful if users would need to register something between 2 authenticators. 

I also think we should avoid a default value here - I'd put `0.0` for a default authenticator & `-1.0` for ldap.

dzikoysk (changes requested):
Thanks for PR. I've left a 2 minor comments :)

dzikoysk (approved):

dzikoysk:
Thanks!

PR #2177, opened 2024-07-27
GH-2177 Extend prometheus with new metrics

solonovamax:
Reworks the prometheus metrics plugin.

Changes:
- Update prometheus metrics to latest version (previously was using the deprecated 0.16.0 simpleclient, updated to version 1.3.1)
- Expose additional jetty metrics
- Expose additional queued thread pool metrics
- Expose additional reposilite-exclusive metrics (summary of response file sizes, failures, uptime, resolved files, deployed files)


reposilite-plugins/prometheus-plugin/src/main/kotlin/com/reposilite/plugin/prometheus/PrometheusPlugin.kt
```diff
@@ -32,38 +41,77 @@ class PrometheusPlugin : ReposilitePlugin() {
...
         )
 
+        JvmMetrics.builder().register()
+        logger.info("Prometheus | JVM metrics has been initialized.")
+
+        val reposiliteMetrics = ReposiliteMetrics.builder {
+            statisticsFacade = facade<StatisticsFacade>()
+            statusFacade = facade<StatusFacade>()
+            failureFacade = facade<FailureFacade>()
+        }.register()
```
dzikoysk, line 56: Is there any reason why we can't just use constructor to pass these values? Using lazy props is a bit dirty and usually it's better to avoid them

reposilite-plugins/prometheus-plugin/src/main/kotlin/com/reposilite/plugin/prometheus/PrometheusPlugin.kt
```diff
@@ -32,38 +41,77 @@ class PrometheusPlugin : ReposilitePlugin() {
...
+            server.handler = handler
+
+            val jettyMetrics = JettyStatisticsHandlerMetrics.builder {
+                statisticsHandler = handler
+            }.register()
+
+            // Abusing request log, as it's the only way to get the time the request was sent
+            server.setRequestLog { request, response ->
+                jettyMetrics.responseTimeSummary.labelValues(response.status.toString())
+                    .observe((System.currentTimeMillis() - request.timeStamp).milliseconds.toDouble(DurationUnit.SECONDS))
```
dzikoysk, line 84: We could probably just get current time in the Javalin's before handler and store it in the ctx.attribute, so later on we can get it back in the after handler. Might be a good idea to add it as a new event to the core impl :thinking:

reposilite-plugins/prometheus-plugin/src/main/kotlin/com/reposilite/plugin/prometheus/PrometheusPlugin.kt
```diff
@@ -32,38 +41,77 @@ class PrometheusPlugin : ReposilitePlugin() {
...
         }
 
         event { event: RoutingSetupEvent ->
-            event.registerRoutes(
-                PrometheusEndpoints(
-                    prometheusFacade = prometheusFacade,
-                    prometheusPath = prometheusPath
-                )
-            )
+            event.registerRoutes(PrometheusEndpoints(prometheusFacade = prometheusFacade, prometheusPath = prometheusPath))
```
dzikoysk, line 111: nit: I don't think it was needed :sweat_smile:

reposilite-plugins/prometheus-plugin/src/main/kotlin/com/reposilite/plugin/prometheus/metrics/QueuedThreadPoolMetrics.kt
```diff
@@ -0,0 +1,66 @@
...
+            QueuedThreadPoolMetrics(config, queuedThreadPool!!).apply {
+                register(registry)
+            }
+    }
+
+    companion object {
+        fun builder(build: Builder.() -> Unit) = builder(PrometheusProperties.get(), build)
+
+        fun builder(config: PrometheusProperties, build: Builder.() -> Unit) = Builder(config).also { it.build() }
+    }
```
dzikoysk, line 65: I don't think these builders or even classes are needed (same for other metrics). I'd rather go for a simple function that just receives everything it needs in the parameters and registers these metrics:

```
fun PrometheusRegistry.registerQueuedThreadPoolMetrics(theadPool) {
   // [...] registers metrics
}
```

dzikoysk (changes requested):
Looks good! Added some minor comments + I think it'd be nice to add at least one test to check if it even starts :sweat_smile: Here's tiny example how to bootstrap it for a plugin: https://github.com/dzikoysk/reposilite/blob/main/reposilite-plugins/migration-plugin/src/test/kotlin/com/reposilite/plugin/migration/MigrationPluginTest.kt#L12

reposilite-plugins/prometheus-plugin/src/main/kotlin/com/reposilite/plugin/prometheus/PrometheusPlugin.kt
solonovamax: tbh the code I copied the metrics shit from was using a builder pattern so I just copied the builder pattern to match it lol

but I can kill the builder if you'd like

reposilite-plugins/prometheus-plugin/src/main/kotlin/com/reposilite/plugin/prometheus/PrometheusPlugin.kt
solonovamax: yeah that would be possible as well

tbh the request does probably have the time, it just needs to be cast to `Request` (yucky)

reposilite-plugins/prometheus-plugin/src/main/kotlin/com/reposilite/plugin/prometheus/PrometheusPlugin.kt
solonovamax: fair, I can revert that

I just try and compress lines when it doesn't increase clarity yk?

reposilite-plugins/prometheus-plugin/src/main/kotlin/com/reposilite/plugin/prometheus/metrics/QueuedThreadPoolMetrics.kt
solonovamax: as mentioned in the other comment, I copied the template for the metrics shit from the impl that comes with the library, as it uses a builder + class pattern

reposilite-plugins/prometheus-plugin/src/main/kotlin/com/reposilite/plugin/prometheus/PrometheusPlugin.kt
dzikoysk: So let's kill all of them with some simple extension functions :+1:

reposilite-plugins/prometheus-plugin/src/main/kotlin/com/reposilite/plugin/prometheus/PrometheusPlugin.kt
dzikoysk: I don't remember, you'd need to jump in with the debugger to investigate all of these underlying properties. Probably just faster to do it on our side for now :sweat_smile:

reposilite-plugins/prometheus-plugin/src/main/kotlin/com/reposilite/plugin/prometheus/PrometheusPlugin.kt
solonovamax: alright, I'll do that later when I'm back at my computer

joshuasing:
This looks great, thank you!

solonovamax:
@dzikoysk I've applied all suggested changes, it's ready to merge now

dzikoysk (approved):

PR #2179, opened 2024-07-29
GH-2120 Switch to using an Server-Sent Event endpoint for Console

### LINKED ISSUE #2120: Replace WebSockets with Server-Sent Events for console
### Request details

Currently, Reposilite's CLI console works by establishing a WebSocket connection between the server and the browser, over which commands and logging lines are exchanged (as seen in https://github.com/dzikoysk/reposilite/blob/229b751a40b111875f51813a8d4615416992b2c1/reposilite-backend/src/main/kotlin/com/reposilite/console/infrastructure/ConsoleWebSocketHandler.kt).

However, this can be troublesome if the Reposilite instance is hosted behind a CDN which does not offer WebSocket-forwarding capabilities. Nearly all other aspects of the frontend work, except for the console. (The URL used for the WS connection is tied to the current window URL, and WebSocket connections do not respect redirects, as that is an optional part of the standard[^1] and it seems major browsers do not follow it by default[^2].)

After raising the topic in the Discord server and chatting with @dzikoysk, it was brought up that the console could be reworked to use [Server-Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) for the log stream (server to client) and the existing `/api/console/execute` endpoint for executing commands (client to server).

As Server-sent Events operate entirely within the HTTP protocol, any compliant HTTP proxy should function correctly without further configuration (which cannot be said of WebSockets, which operate under the `ws` protocol).

Javalin has a documentation page about the use of Server-sent Events: https://javalin.io/documentation#server-sent-events.

The existing WebSocket-powered endpoint, `/api/console/sock`, and its associated code would be retained for backwards compatibility, perhaps for removal in 4.x.

[^1]: "If the status code received from the server is not 101, the client handles the response per HTTP [[RFC2616](https://www.rfc-editor.org/rfc/rfc2616)] procedures.  In particular, the client might perform authentication if it receives a 401 status code; the server might redirect the clien using a 3xx status code (but clients are not required to follow them), etc." ~ [RFC 6445: The WebSocket Protocol](https://www.rfc-editor.org/rfc/rfc6455)
[^2]: My testing shows the redirect is not followed by Chrome and [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1052909).

granny:
This PR adds an SSE endpoint to reposilite for relaying log messages to the CLI console. The `/api/console/execute` endpoint is used to send commands. Closes #2120

## TODO/More Info Needed

The checklists below are a mixture of talking points that I need more input on as well as "TODOs" that need to be resolved. Please excuse the mess..

### Meta

- [x] ~`/api/console/log` is the endpoint. I'm open to ideas if you'd like to call it something else.~ It's good enough for now.
- [x] ~There's not really a OpenAPI Spec for SSE endpoints so I put some basic info for now.~
- [ ] Currently all event messages are sent under the "log" event. Honestly it's kind of repetitive with the endpoint also being "log", so it'd make sense to send all messages under the default "message" event if the endpoint ends up staying as `/log`.
- [ ] Should there be special formatting of the `/api/console/execute` endpoint if it's done through CLI? Currently it just outputs as `REMOTE EXECUTION /api/console/execute from 127.0.0.1` since right now there's no way to differentiate between a remote execution and what was sent through CLI.

### Backend

- [x] ~`Failed to close writer (java.lang.InterruptedException)` shows up when stopping the application.~ Unrelated to my PR.
- [x] ~`SseClient` connections aren't properly closed when stopping the application.~ All connections are closed in the `ReposiliteDisposeEvent` event under ConsolePlugin.
- [x] ~`SseClient` connections aren't properly closed in general. They only close when an attempt is made to push an event (which throws a logged exception)~ All connections are sent a "ping" comment every second to verify that they are still connected. Javalin will handle closing the connection for us if the ping comment does not get received.

### Frontend

- [x] ~Adds an extra dependency, [extended-eventsource](https://github.com/lukas-reining/eventsource), which reimplements [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) using [Fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). This is required to pass an authentication header to the SSE endpoint, since you can't do so with the built-in EventSource API.~ Seems to work fine.
- [x] ~Closing the browser tab or refreshing the page doesn't properly close the SSE connection. The backend fails to notice that there is no longer a connection, and ends up throwing a small CME error the next time a console command is sent from the same credentials.~ Closing the browser tab or refreshing the page properly closes the SSE connection now.
- [x] ~`Console connection error - Make sure that WebSockets are enabled` needs to be changed/reworded~ It has been adjusted.
- [x] ~Stop using built-in EventSource for readystate constants (the ones from extended-eventsource return undefined and I don't know why)~ Good to leave as is.
- [x] ~~Currently if the SSE connection fails/errors/disconnects, then it doesn't attempt to reconnect. Should it? How many times should it try to reconnect before giving up?~~ It won't attempt to reconnect.

### Other

I took the liberty of renaming the internal `CliEndpoint` class to `ConsoleWebSocketHandler`, to keep file names consistent. I can revert this if it was left like that on purpose.


granny:
## Backend changes

- All SSE connections are now closed when shutting down reposilite. 
- All SSE connections now get pinged every second (Javalin closes any connections where the ping fails)

## Frontend changes

- Close sse connection before the page is reloaded/refreshed. This is technically not needed, but it stops an annoying error from appearing in the browser console.

## Etc

I believe I've fixed all the errors/bugs. This PR is now ready for review.

reposilite-backend/src/main/kotlin/com/reposilite/console/application/ConsolePlugin.kt
```diff
@@ -35,13 +36,28 @@ import com.reposilite.plugin.reposilite
...
+            forwardedIp = sharedConfigurationFacade.getDomainSettings<WebSettings>().computed { it.forwardedIp }
+        )
+
+        val watcher = reposilite().scheduler.scheduleWithFixedDelay({
+            // use an iterator instead of forEach to avoid CME
+            val iterator = client.users.iterator()
+            while (iterator.hasNext()) {
+                iterator.next().key.sendComment("ping")
+            }
+        }, 0, 1, TimeUnit.SECONDS)
```
dzikoysk, line 60: I don't think that such frequent ping is mandatory here, so maybe we can increase the interval a little bit? For WS we used 5s, so that could be an option

reposilite-backend/src/main/kotlin/com/reposilite/console/application/ConsolePlugin.kt
```diff
@@ -66,14 +82,19 @@ internal class ConsolePlugin : ReposilitePlugin() {
...
                         accessTokenFacade = facade(),
                         authenticationFacade = facade(),
                         consoleFacade = consoleFacade,
                         forwardedIp = sharedConfigurationFacade.getDomainSettings<WebSettings>().computed { it.forwardedIp }
                     )
                 )
+                it.sse(
+                    // TODO: does this need better endpoint name?
+                    "/api/console/log",
+                    client
```
dzikoysk, line 96: I think `"/api/console/log"` is okay, in the end this endpoint only returns live logs from the cli now

reposilite-backend/src/main/kotlin/com/reposilite/console/application/ConsolePlugin.kt
```diff
@@ -35,13 +36,28 @@ import com.reposilite.plugin.reposilite
...
+            authenticationFacade = facade(),
+            forwardedIp = sharedConfigurationFacade.getDomainSettings<WebSettings>().computed { it.forwardedIp }
+        )
+
+        val watcher = reposilite().scheduler.scheduleWithFixedDelay({
+            // use an iterator instead of forEach to avoid CME
+            val iterator = client.users.iterator()
+            while (iterator.hasNext()) {
+                iterator.next().key.sendComment("ping")
+            }
```
dzikoysk, line 59: I'm not sure if keeping the state in sse handler and then processing it in the plugin is a good idea. Maybe we should create SseService that keeps some state & exposes API to interact with it.

I also think it'd be better to spawn ping scheduler for each new connection instead of keeping it at the global level. Then, you could stop the task by regisdtering `sse.onClose {}` hook.

reposilite-backend/src/main/kotlin/com/reposilite/console/application/ConsolePlugin.kt
```diff
@@ -91,6 +112,10 @@ internal class ConsolePlugin : ReposilitePlugin() {
 
         event { _: ReposiliteDisposeEvent ->
             consoleFacade.commandExecutor.stop()
+            client.users.forEach {
+                it.key.close()
+            }
+            watcher.cancel(false)
```
dzikoysk, line 118: Related to what I said before - it'd be nice mange the state in the service class.

And watcher should be probably disabled before closing the sse clients :thinking:

reposilite-backend/src/main/kotlin/com/reposilite/console/infrastructure/ConsoleSseHandler.kt
```diff
@@ -0,0 +1,112 @@
...
+    private val journalist: ReposiliteJournalist,
+    private val accessTokenFacade: AccessTokenFacade,
+    private val authenticationFacade: AuthenticationFacade,
+    private val forwardedIp: Reference<String>
+) : Consumer<SseClient> {
+
+    internal val users: WeakHashMap<SseClient, SseSession> = WeakHashMap()
+
+    @OpenApi(
+        // TODO: does this need better endpoint name?
```
dzikoysk, line 42: :white_check_mark:

reposilite-backend/src/main/kotlin/com/reposilite/console/infrastructure/ConsoleSseHandler.kt
```diff
@@ -0,0 +1,112 @@
...
+        path = "/api/console/log",
+        methods = [HttpMethod.GET],
+        // TODO: the responses parameter is *technically* not really needed, but would be nice to have
```
dzikoysk, line 45: I think it's good to have it here as some sort of docs for this endpoint :+1:

reposilite-backend/src/main/kotlin/com/reposilite/console/infrastructure/ConsoleSseHandler.kt
```diff
@@ -0,0 +1,112 @@
...
+        responses = [
+            OpenApiResponse(
+                status = "200",
+                // TODO: better description?
+                description = "Continuously sends out the log as data messages",
```
dzikoysk, line 50: I'd mention that the SSE endpoint, despite the log, also produces ping values as comments

reposilite-backend/src/main/kotlin/com/reposilite/console/infrastructure/ConsoleSseHandler.kt
```diff
@@ -0,0 +1,112 @@
...
+                content = [OpenApiContent(from = ExecutionResponse::class)]
+            )
+        ],
+        tags = ["Console"]
+    )
+    override fun accept(sse: SseClient) {
```
dzikoysk, line 56: Let's call it `handleSseLiveLog`, I'm not a big fan of using these generic interfaces

reposilite-backend/src/main/kotlin/com/reposilite/console/infrastructure/ConsoleSseHandler.kt
```diff
@@ -0,0 +1,112 @@
...
+        sse.keepAlive()
+        sse.onClose {  ->
+            val session = users.remove(sse) ?: return@onClose
+            journalist.logger.info("CLI | ${session.identifier} closed connection")
+            journalist.unsubscribe(session.subscriberId)
+        }
+
+        when (users[sse]) {
+            null ->
```
dzikoysk, line 65: I think it is redundant, each SseClient instance will be treated as a different value, so it'll be always null for a new connection.

reposilite-backend/src/main/kotlin/com/reposilite/console/infrastructure/ConsoleSseHandler.kt
```diff
@@ -0,0 +1,112 @@
...
+                    .peek { identifier ->
+                        journalist.logger.info("CLI | $identifier accessed remote console")
+
+                        val subscriberId = journalist.subscribe {
+                            // TODO: do better
+                            if (!sse.terminated()) {
+                                sse.sendEvent(SSE_EVENT_NAME, it.value)
+                            } else {
+                                sse.close()
+                            }
```
dzikoysk, line 76: It is not needed, if sse channel is terminated, Javalin automatically closes the client:

* https://github.com/javalin/javalin/blob/master/javalin/src/main/java/io/javalin/http/sse/SseClient.kt#L68-L70

reposilite-backend/src/main/kotlin/com/reposilite/console/infrastructure/ConsoleSseHandler.kt
```diff
@@ -0,0 +1,112 @@
...
+            .flatMap { authenticationFacade.authenticateByCredentials(it) }
+            .filter(
+                { accessTokenFacade.hasPermission(it.identifier, AccessTokenPermission.MANAGER) },
+                { unauthorized("Unauthorized CLI access request from ${connection.getHost()}") }
+            )
+            .map { "${it.name}@${connection.getHost()}" }
+    }
+
+    private fun Context.getHost(): String =
+        header(forwardedIp.get()) ?: req().remoteAddr
```
dzikoysk, line 111: You can use `ctx.ip()`, it's dynamically resolved value that already does the same thing :+1: 

* Our cfg: https://github.com/dzikoysk/reposilite/blob/main/reposilite-backend/src/main/kotlin/com/reposilite/web/application/JavalinConfiguration.kt#L80

reposilite-frontend/src/components/console/ConsoleView.vue
```diff
@@ -60,7 +60,8 @@ const setupConnection = () => {
   }
   onError.value = error => {
     console.log(error)
-    createToast(`Console connection error - Make sure that WebSockets are enabled.`, { type: 'danger' })
+    // TODO: change error content
+    createToast(`Console connection error - ???.`, { type: 'danger' })
```
dzikoysk, line 64: Maybe just `Console connection error - Cannot establish SSE connection`

reposilite-frontend/src/store/console/connection.js
```diff
@@ -16,17 +16,23 @@
...
+const { client } = useSession()
 
 const connection = ref()
 const command = ref("")
 
 export default function useConsole() {
-  const consoleAddress = createURL("/api/console/sock")
-    .replace("https", "wss")
-    .replace("http", "ws")
+  // TODO: does this need better endpoint name?
```
dzikoysk, line 28: :white_check_mark:

reposilite-frontend/src/store/console/connection.js
```diff
@@ -16,17 +16,23 @@
...
+  const consoleAddress = createURL("/api/console/log");
 
-  const isConnected = () =>
-    connection.value?.readyState === WebSocket.OPEN
+  const isConnected = () => {
+    // TODO: stop using built-in EventSource for readystate constants
+    //  (the ones from extended-eventsource return undefined for some reason)
```
dzikoysk, line 33: I think it's fine to use the EventSource value if it just works

reposilite-frontend/src/store/console/connection.js
```diff
@@ -67,37 +73,44 @@ export default function useConsole() {
     command.value = history.value[historyIdx.value]
   }
 
+  // this is needed to stop an error from appearing in console when
+  // switching/refreshing the page without closing the connection
+  // TODO: move somewhere else?
+  window.onbeforeunload = function () {
+    close();
+  };
```
dzikoysk, line 81: I'd do it in the `onopen` listener, but despite that I think it's fine to have it - just keep this comment to keep it transparent on why it's needed.

reposilite-frontend/src/store/console/connection.js
```diff
@@ -67,37 +73,44 @@ export default function useConsole() {
...
 
   const connect = (token) => {
     try {
-      connection.value = new WebSocket(consoleAddress)
+      connection.value = new Eventsource(consoleAddress, {
+        headers: {
+          Authorization: `xBasic ${btoa(`${token.name}:${token.secret}`)}`
+        },
+        // TODO: should we try to reconnect? how many attempts should be made?
+        disableRetry: true
```
dzikoysk, line 95: As far as I see, the SSE disconnects on errors. Retries help to recover from that, so I think it shouldn't be needed in our case then :thinking:

dzikoysk (changes requested):
Awesome, thank you! I've left some minor comments regarding the implementation :heart_hands:

reposilite-backend/src/main/kotlin/com/reposilite/console/application/ConsolePlugin.kt
granny: I completely agree with both statements. That's actually what I wanted to do originally, but settled on this just to get a working PoC out. How would you like the SseService to be implemented? This is the first time I've messed with kotlin & the reposilite internals so I might need a bit of hand holding :P

reposilite-frontend/src/store/console/connection.js
granny: I'm a bit confused by the wording of this. Do I leave it as is and just remove the TODO comment?

reposilite-backend/src/main/kotlin/com/reposilite/console/application/ConsolePlugin.kt
dzikoysk: Yeah, looks like a workaround to some underlying issue. It's logged without formatting, so I guess the error happens insde the logging impl - I've found regular HashMap being used here:

* https://github.com/reposilite-playground/journalist/blob/main/journalist/src/main/java/com/reposilite/journalist/backend/PublisherLogger.java#L14

So it might be a good idea to fix it here and see if it helps. In the meantime, let's keep a "proper" approach here (5s) and ignore the error.

reposilite-frontend/src/store/console/connection.js
dzikoysk: I was just replying to the question from todo :sweat_smile: I think it's okay to disable retries and you can remove the comment :ok_hand:

reposilite-backend/src/main/kotlin/com/reposilite/console/application/ConsolePlugin.kt
dzikoysk: It was one of the first comments in this code review, but now I think that after applying other changes and spawning a new task per each connection in the SseHandler itself, we may actually not need a dedicated service for that, because whole request lifecycle will be covered by the handler.

reposilite-backend/src/main/kotlin/com/reposilite/console/application/ConsolePlugin.kt
dzikoysk: I've just pushed a newer version of Journalist, you can try to bump it to 1.0.12 to check if it helped.

granny: That fixed the CME errors :)

reposilite-backend/src/main/kotlin/com/reposilite/console/application/ConsolePlugin.kt
granny: I've moved the watcher into the SSE Handler

dzikoysk:
>  Currently all event messages are sent under the "log" event. Honestly it's kind of repetitive with the endpoint also being "log", so it'd make sense to send all messages under the default "message" event if the endpoint ends up staying as /log.

I think that's fine - it costs us nothing and and this endpoint will be most likely used only by our internal frontend implementation.

>  Should there be special formatting of the /api/console/execute endpoint if it's done through CLI? Currently it just outputs as REMOTE EXECUTION /api/console/execute from 127.0.0.1 since right now there's no way to differentiate between a remote execution and what was sent through CLI.

That would be a great enhancement, I guess we could add some sort of `ExecutionSource` enum that would indicate if command was invoked from cli/api/web ui. It's up to you if you address that, we can even move this to separate PR to not introduce more changes in the scope of this one :)

reposilite-backend/src/main/kotlin/com/reposilite/console/application/ConsolePlugin.kt
dzikoysk: Nice! :)

dzikoysk (approved):

dzikoysk:
Thanks, great work! :heart_hands:

PR #2198, opened 2024-08-17
Fix broken link

CoPokBl:
Fix https://reposilite.com/guide/general "Standalone" link taking you to `/standalone` instead of `/jar` which it should.


joshuasing (approved):

dzikoysk:
Thanks!

Issue #2222, opened 2024-09-10
Documentation on Docker missing line break character.


TOTHT0MI:
### What happened?

The docker installation documentation external config part is missing the line break character '\\' on line 4 and 5.
This way when copy pasting the command, it fails to execute.

See External Configuration: [https://reposilite.com/guide/docker](https://reposilite.com/guide/docker)
![image](https://github.com/user-attachments/assets/e4e9efef-8adb-462d-919a-594d5b38dc6c)

### Reposilite version

3.x

### Relevant log output

```shell
-
```

PR #2223, opened 2024-09-14
GH-2222 Added line breaks to docker command

TOTHT0MI:
Added line break char ('\\') to docker run command. Copy pasting the command will now work instead of failing.


dzikoysk (approved):

dzikoysk:
Thanks!

Issue #2234, opened 2024-09-29
Repository snippets frontend inconsistencies


jaqobb:
### What happened?

I have noticed a few inconsistencies in repository snippets. Mainly:
* Gradle Kotlin & Gradle Groovy repository snippets do not have the repository name included. As an example, trying to copy the following Gradle Kotlin repository to the clipboard:
```kotlin
maven {
    name = "jaqobbRepositoryReleases"
    url = uri("https://repository.jaqobb.dev/releases")
}
```
results in the following being copied:
```kotlin
maven {
    url = uri("https://repository.jaqobb.dev/releases")
}
```
* Maven and SBT repository snippets have different amount of spaces / tab length (2) compared to Gradle Kotlin & Gradle Kotlin (4). While I do not use Scala and cannot speak on that, I understand there are people using 2 spaces in XML files. As such, I believe the amount of spaces should be updated to be the same (2/4) across all repository snippets or a simple configuration should be added allowing this to be adjusted by the end-user.

I am not really proficient in frontend technologies and cannot make a PR for that. However, I am pretty sure these do not require any huge changes.

### Reposilite version

3.x

WithoutAName25:
I have fixed the first issue with the inconsistencies between the displayed text and the copied text. It would be easy to change the indent of the snippets if that is wanted. If the indent should be configurable it would be more complex but I can take a look at it if this is desired.

PR #2248, opened 2024-10-02
GH-2234 Fix snippet inconsistencies

WithoutAName25:
The snippet is now extracted from the HTML via textContent and not generated separately to avoid inconsistencies between the displayed text and the copied text


dzikoysk (approved):

dzikoysk:
Thanks, looks good (and works :sweat_smile:)! :heart:

Issue #2251, opened 2024-10-08
`RepositoryService.findInputStream` does not emit `ResolvedFileEvent`


solonovamax:
### What happened?

The `ResolvedFileEvent` event is not emitted with the `findInputStream` method, thus causing the reposilite prometheus metrics to not be updated properly:
```
# HELP reposilite_resolved_total Total resolved files count
# TYPE reposilite_resolved_total counter
reposilite_resolved_total 0.0
# HELP reposilite_response_file_size_bytes Size in bytes of response files
# TYPE reposilite_response_file_size_bytes summary
reposilite_response_file_size_bytes{quantile="0.01"} NaN
reposilite_response_file_size_bytes{quantile="0.05"} NaN
reposilite_response_file_size_bytes{quantile="0.1"} NaN
reposilite_response_file_size_bytes{quantile="0.5"} NaN
reposilite_response_file_size_bytes{quantile="0.9"} NaN
reposilite_response_file_size_bytes{quantile="0.95"} NaN
reposilite_response_file_size_bytes{quantile="0.99"} NaN
reposilite_response_file_size_bytes_count 0
reposilite_response_file_size_bytes_sum 0.0
```
(note: a file *has* been requested. I'm assuming that the `findInputStream` method is used to return the response.)

The `ResolvedFileEvent` event would probably be best emitted as part of the `recordResolvedRequest` method.
Additionally, `recordResolvedRequest` does not seem to be invoked when `RepositoryService.findInputStream` is invoked, either.

### Reposilite version

3.x

### Relevant log output

```shell
N/A
```

dzikoysk:
The `ResolvedFileEvent` is supposed to contain `DocumentInfo` handled by `findFile`, while `findInputStream`  provides only raw data. I think we should introduce a new (generic) event to represent a simple resolution that will cover both of them.

solonovamax:
> The `ResolvedFileEvent` is supposed to contain `DocumentInfo` handled by `findFile`, while `findInputStream` provides only raw data. I think we should introduce a new (generic) event to represent a simple resolution that will cover both of them.

Since the `ResolvedFileEvent` is only meant to be emitted when the file details are queried, I'm making a PR that introduces a `FileResponseEvent`, specifically for when a file is returned via an http request.

solonovamax:
actually nvm I think it might be good to just drop the `reposilite_response_file_size_bytes` all together, as you can just use `jetty_response_bytes` instead?

[...]

PR #2258, opened 2024-10-30
GH-2251 Fix Jetty Metrics Recording

solonovamax:
Hopefully this uncooks the jetty metrics

Also bumps the version for prometheus metrics.

### DIFF
--- a/reposilite-plugins/prometheus-plugin/build.gradle.kts
+++ b/reposilite-plugins/prometheus-plugin/build.gradle.kts
@@ -28,10 +28,10 @@ application {
 dependencies {
     compileOnly(project(":reposilite-backend"))
 
-    val prometheus = "1.3.1"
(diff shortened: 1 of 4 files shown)


dzikoysk:
> Hopefully

Did you try to run this? :sweat_smile:

solonovamax:
> > Hopefully
> 
> Did you try to run this? 😅

you never know, some magic issue might pop up later

solonovamax:
actually I think I did something dumb, is there an easy way to take an arbitrary string and normalize it as a url path?

eg.

`//metrics///` -> `/metrics`?
I couldn't find any from a quick look earlier

dzikoysk:
You could try with our `Location` class I think, it's intended to clean up messed paths _(in general)_

dzikoysk (approved):

dzikoysk:
Thanks, let's see how it'll work :shipit:

Issue #2282, opened 2024-11-23
Add support for Major and Major.Minor Docker tags


kirbylink:
### Request details

**Problem Statement**  
Currently, the Docker images published for Reposilite are tagged only with `latest` and the full version (e.g., `3.5.19`). While this works for many use cases, it can make managing updates more challenging for users who want to:
- Restrict updates to a specific major version (e.g., `3.x`) to avoid breaking changes.
- Apply updates only within a specific minor version (e.g., `3.5.x`) for added stability.

This lack of additional tags can lead to either using the full version tag (requiring frequent manual updates) or relying on `latest`, which carries the risk of breaking changes.

**Proposed Solution**  
Enhance the tagging process for Docker images by adding the following additional tags:
- `Major` (e.g., `3`)
- `Major.Minor` (e.g., `3.5`)

These tags would complement the existing `latest` and full release tags, giving users more flexibility to manage updates in their environments.

**Benefits**  
- Users can opt-in to update strategies that better suit their requirements:
  - Use the `Major` tag to stay on the latest version within a major release.
  - Use the `Major.Minor` tag for stability within a minor release series.
- Reduced risk of unexpected breaking changes when relying on the `latest` tag.

**Implementation Details**  
In the `docker` job of the GitHub Actions workflow, the additional tags can be generated by extracting the `Major` and `Minor` parts of the version (e.g., from a variable or computed from the full version). These tags would then be passed to the `docker/build-push-action` step alongside the existing tags.

**Additional Context**  
I am happy to open a PR to implement this feature, as I already have a working solution. Please let me know if this aligns with the goals of the project or if any adjustments are needed before proceeding.

dzikoysk:
I think it'd be a great improvement, feel free to submit the PR then :shipit:

PR #2285, opened 2024-11-23
GH-2282 Add Docker tags for Major and Major.Minor versions

kirbylink:
**Purpose**  
This pull request enhances the Docker image tagging process by adding additional tags for `Major` and `Major.Minor` versions alongside the existing `latest` and full release version tags.  

**Why this change?**  
Using only `latest` or the full version (e.g., `3.5.19`) for Docker tags can make managing updates more challenging, especially for users who want to restrict updates to specific major or minor versions. For example:
- A tag like `3.5` allows users to get updates within the `3.5.x` series while avoiding potential changes in version `3.6` without explicit changing the tag.
- Similarly, a `3` tag ensures updates only within the `3.x` series to avoid auto updates with breaking changes.

Adding these tags provides users with more flexibility and control over updates in their environments.

**What was changed?**  
- The `docker` job in the GitHub Actions workflow now generates the following tags for Docker images:
  - `latest`
  - `Major.Minor` (e.g., `3.5`)
  - `Major` (e.g., `3`)
  - Full release version (e.g., `3.5.19`)
- These tags are pushed to both DockerHub and GitHub Container Registry (GHCR).

**How does it work?**  
- A new step (`Generate Docker Tags`) extracts the `Major` and `Minor` parts of the version from `${{ needs.github.outputs.version }}`.
- The generated tags are passed to the `docker/build-push-action`, ensuring all tags are applied to the built image and pushed to the registries.

**Testing**  
While testing was limited due to the lack of actual registry credentials, the logic for tag generation was validated independently to ensure correctness. 

Fixing #2282

It might be possible to update parts of the documentation, but I wasn't sure if and where it would make sense.

Additional notes:
During my tests with GitHub workflows i saw the following warning:
> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

I also used `set-output` because it is already used in the workflow.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -101,6 +101,15 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: "Generate Docker Tags"
+        id: generate-tags
+        run: |
+          VERSION="${{ needs.github.outputs.version }}"
+          MAJOR=$(echo $VERSION | cut -d '.' -f 1)
+          MINOR=$(echo $VERSION | cut -d '.' -f 2)
+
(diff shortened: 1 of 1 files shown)


dzikoysk (approved):

dzikoysk:
Looks good to me, I think we can just merge it and see how it goes. In case of any errors, we can address it later in separate PRs :)

Issue #1762, opened 2023-03-30
Image requires to be run as root and does not support random uids


bo0ts:
### What happened?

The docker image requires to be run as root since it adds a new user and group. This makes it hard to run the image as-is on OpenShift/OKD or in constrained environments.

Supporting random UUIDs would be a great feature.

### Reposilite version

3.x

dzikoysk:
I don't have much knowledge about OpenShift/OKD, but current setup makes sure that Reposilite is not launched as root user inside the container (the default behavior of Docker that is our current target). Do you have any recommendations about possible improvements to the existing scripts?

* https://github.com/dzikoysk/reposilite/blob/main/Dockerfile
* https://github.com/dzikoysk/reposilite/blob/main/entrypoint.sh

Maybe we should add some kind of new env variable that would launch Reposilite directly, without custom user 🤔

bo0ts:
As far as I can see the image currently needs to be started as root to perform `adduser` and `addgroup`. It only switches into the user afterwards. Usually `adduser`/`addgroup`/`chown` and the switch to the new user are done during the image build.

To make the image run with arbitrary user ids the OpenShift doc (https://docs.openshift.com/container-platform/4.9/openshift_images/create-images.html#images-create-guide-openshift_create-images) says 

> For an image to support running as an arbitrary user, directories and files that are written to by processes in the image must be owned by the root group and be read/writable by that group. Files to be executed must also have group execute permissions.

So adding `chgrp -R 0 /app/ && chmod -R g+rwX /app/ && chgrp -R 0 /var/log/reposilite && chmod -R g+rwX /var/log/reposilite` should do the trick.

Unfortunately I haven't been able to build the image locally with podman on an M1 MacBook yet, so I cannot test this right now.

dzikoysk:
I'd like to keep just one image that covers all we need. My suggestion was to do sth like:

```bash
#!/bin/bash

REPOSILITE_ARGS=$REPOSILITE_OPTS
case "$REPOSILITE_OPTS" in
  *"--working-directory"*) ;;
  *"-wd"*                ) ;;
  *                      ) REPOSILITE_ARGS="--working-directory=/app/data $REPOSILITE_ARGS";;
esac

# // if (REPOSILITE_WITHOUT_RUNUSER):
#   java -jar [...]
# // else:
GROUP_ID="${PGID:-999}"
addgroup --gid $GROUP_ID reposilite

USER_ID="${PUID:-999}"
[...]

Issue #1657, opened 2022-12-14
Error deploying in OpenShift because of chown


bennyevangelista:
### What happened?

Hi, I'm trying to setup a POC of Reposilite for my organization, it runs perfectly via podman on my development machine,  but when I try to deploy it in an openshift cluster the pod crashes right after the start. Below you will find the log entries before the crash, and here's my deployment yaml:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: dep-reposilite
spec:
  replicas: 1
  selector:
    matchLabels:
      deployment: reposilite
  template:
    metadata:
      labels:
        deployment: reposilite
    spec:      
      containers:      
      - name: reposilite
        image: dzikoysk/reposilite:3.2.0       
        ports:
        - containerPort: 8080
          name: tcp-8080
          protocol: TCP
        resources:
          limits:
            cpu: 500m
            memory: 256M  
```   

Does the container need root priviliges to run? In this case this would break one of my company's security policy.

Thanks for the support!

### Reposilite version

3.x

### Relevant log output

```shell
chown: changing ownership of '/app/data/lost+found': Operation not permitted
chown: changing ownership of '/app/data': Operation not permitted
chown: changing ownership of '/app/entrypoint.sh': Operation not permitted
chown: changing ownership of '/app/reposilite.jar': Operation not permitted
chown: changing ownership of '/app': Operation not permitted
chown: changing ownership of '/var/log/reposilite': Operation not permitted
runuser: may not be used by non-root users
```

dzikoysk:
Hey :) Root is needed, so container can create dedicated user for Reposilite and run application as non root user:

* https://github.com/dzikoysk/reposilite/blob/main/entrypoint.sh

By default application would run as root, so this is exactly needed to launch application as non root user.

bennyevangelista:
Thanks for the quick reply :) Maybe I'm missing something, but what's the need for a dedicated user? I mean, as a java application it should be sufficient for the container to run command "java -jar ...". I don't have experience with kubernetes, but in Openshift containers run as a non-root unique user, separated from other system users. Due to this, I'm afraid I won't be able to try Reposilite in my environment.

Thanks :)

dzikoysk:
It's a good practice to not run application inside container as root user and usually such software creates dedicated user, see:

* https://github.com/dzikoysk/reposilite/issues/1200
* https://github.com/dzikoysk/reposilite/issues/1634

[...]

Issue #1634, opened 2022-11-09
Run docker image as a given user (uid/gid)


hereisderek:
### Request details

is there a way to run the docker as a specific user (with given uid/gid) instead of the hard coded 999:999 ?  

one of the common practices is to pass in as PUID and PGID environment variable

Thanks

dzikoysk:
Not really, those two are hardcoded in Docker image file: 

* https://github.com/dzikoysk/reposilite/blob/864d2d013334846e6f795bdf5437f67a5bbe1ac3/Dockerfile#L38

As far as I remember quite a lot of popular projects use this 999:999 combination, why would you want to change that?

dzikoysk:
I've used recommended by you variable names - `PUID` & `PGID`. The script will create user & group at the start of the container.

Issue #1200, opened 2022-03-28
Docker container running as root


larsw:
### What happened?

It's a well-known fact that running Docker container as UID=0/root is a security threat.
A consequence of running as non-root, is that you can no longer bind to ports < 1024.

### Reposilite version

3.x

PR #2289, opened 2024-11-24
[STEP 2] Migrate Docker images to Alpine and BellSoft Liberica

### LINKED ISSUE #2288: Deprecation of openJDK docker images
As of December 20th 2022 the openJDK images from the docker library were deprecated (see docker-library/openjdk@d2f1e2d  and docker-library/openjdk#505). With this comes with a handful of questions for continuation of the reposilite docker images. Below are some choices to go with moving forward.

## One base image to rule them all
Pick a new single base image for future reposilite images
- Amazon's Corretto JVM images
  - `amazoncorretto:${JDK}-al2023` based off of Amazon Linux 2023
  - `amazoncorretto:${JDK}-alpine` based off Alpine Linux
- Eclipse's Temurin JVM images
  - `eclipse-temurin:${JDK}` based off Ubuntu
  - `eclipse-temurin:${JDK}-alpine` based off Alpine Linux
  - `eclipse-temurin:${JDK}-windowsservercore` based off Windows Server 2016 (I advise against this one)
- BellSoft Liberica JVM images
  - `bellsoft/liberica-runtime-container:jre-${JDK}`
- IBM Semeru (OpenJ9) JVM images
  - `ibm-semeru-runtimes:open-${JDK}` based off Ubuntu
- SapMachine JVM images
  - `sapmachine:${JDK}` based off Ubuntu

With one of these images rework the Dockerfile to use said image.
[Example](https://gist.github.com/tech-6/0ea86682c2dbd922252a375c199413e7)

## How about all of them
Publish a build for x amount of base runtime images using artifacts from a build image. Allowing users to choose their preferred JVM runtime to run their repo with. A method of doing this could be using `docker buildx bake` to make a build job to build the JAR file and cache them for building the runtime images (which could be done in parallel).

---

More commentary is wanted or more potential solutions. Many thanks.

tech-6:
- Default user and group id is now 977 as 999 was taken.
> [!CAUTION]
> **!!This may cause permissions issues in existing installations**. 
> The above mentioned permissions issues can be fixed by either:
> - mounting the volume with another container as a root user and changing ownership of `/app/data` (including their contents) to 977:977
> - `chown -R 977:977 /var/lib/docker/volumes/${VOLUME_NAME_OR_HASH}` of the volume from the host's volume directory.
> - `chmod -R  777 /app/data` (don't do this)
- Dockerfile and entrypoint.sh was changed to run NO application code as a privileged user. When the container is running there is no code executed in a privileged context, as anything requiring privileges is run during build time. 
- Addition of a .dockerignore file to ignore non-application files and improve docker cache hits.

Proposed fix #2288

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,17 @@
+# Per developer files
+build/
+.idea/
+.gradle/
+.kotlin/
+
+# Dockerfile related files do not need to be included
+Dockerfile
+docker-compose.yml
+.env
+
+# Not used in running application
+reposilite-site
+reposilite-test
+reposilite-plugins
+.github/
+.codecov.yml
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax=docker.io/docker/dockerfile:1.7-labs
 
 # Build stage
-FROM bellsoft/liberica-runtime-container:jdk-21-cds-musl AS build
+FROM bellsoft/liberica-runtime-container:jdk-21-musl AS build
 COPY --exclude=entrypoint.sh . /home/reposilite-build
 WORKDIR /home/reposilite-build
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,20 @@
+# syntax=docker.io/docker/dockerfile:1.7-labs
+
 # Build stage
-FROM openjdk:21-slim AS build
-COPY . /home/reposilite-build
+FROM bellsoft/liberica-runtime-container:jdk-21-cds-musl AS build
+COPY --exclude=entrypoint.sh . /home/reposilite-build
 WORKDIR /home/reposilite-build
-RUN \
-  rm -rf reposilite-frontend/node_modules
-RUN \
-  apt-get update; apt-get install -y curl
-RUN \
-  export GRADLE_OPTS="-Djdk.lang.Process.launchMechanism=vfork" && \
-  chmod +x gradlew && \
-  bash gradlew :reposilite-backend:shadowJar --no-daemon --stacktrace
+
+# Get build dependencies seperately so they can cache
+RUN <<EOF
+   apk --no-cache add nodejs
+EOF
+
+# The below line will show an Error in some IDE's, It is valid Dockerfile.
+RUN --mount=type=cache,target=/root/.gradle <<EOF
+  export GRADLE_OPTS="-Djdk.lang.Process.launchMechanism=vfork"
+  ./gradlew :reposilite-backend:shadowJar --no-daemon --stacktrace
+EOF
 
 # Build-time metadata stage
 ARG BUILD_DATE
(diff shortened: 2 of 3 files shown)


tech-6:
Seems ready for review, has been stable in my testing environment.

dzikoysk:
Hmm... I'm a bit worried about changing current behaviour - we have quite wide variety of users, even on some exotic setups. That's also the reason why we basically moved a lot of the logic to the entrypoint.sh script.

I think it requires deeper investigation, I'll try to check some historical issues. I guess you could also browse them by searching for Dockerfile or entrypoint keywords.

tech-6:
The main issues that came to mind are #1762, #1200, #1634. 
- #1200 would continue to be resolved as we would be running as non root.
- #1657, #1762 Pertain to opinionated K8s distributions, more research needed to determine compatibility

Below is a todo list on testing deployments,

- [x] `docker-compose.yml`
- [ ] Kubernetes (I'll most likely try it on my home "cluster")
- [ ] Redhat Openshift re: #1657 #1762

Could be helpful to see if the users from #1657 and #1762 if they got it up and running and for potential testing.

tech-6:
Confirmed a working fresh install on RedHat Openshift (see above link)
image
Seems like Openshift's UID randomization still works with user 977 and the rootless container.

solonovamax:
A few questions:
- Why was a liberica jdk chosen over a jdk from eclipse temurin (eg. `eclipse-temurin:21-jre-alpine`)? What is the benefit of this? Looking at https://whichjdk.com/, there does not seem to be any significant benefit, so I'm unsure why it was chosen.
- Why was a docker image with [Class Data Sharing](https://docs.oracle.com/en/java/javase/22/vm/class-data-sharing.html) chosen for the build stage, even though it is never enabled/used?

dzikoysk:
> Why was a liberica jdk chosen over a jdk from eclipse temurin (eg. eclipse-temurin:21-jre-alpine)? What is the benefit of this? Looking at https://whichjdk.com/, there does not seem to be any significant benefit, so I'm unsure why it was chosen.

1. It's smaller
- `eclipse-temurin:21-jre-alpine` - 71mb
- `liberica-runtime-container:jre21-slim-musl` - 45mb
- `liberica-runtime-container:jre-21-cds-slim-musl` - 52mb
2. I didn't know they have JRE builds

One nice thing about temurin is that _maybe_ we could avoid changing user & group ids.

> Why was a docker image with [Class Data Sharing](https://docs.oracle.com/en/java/javase/22/vm/class-data-sharing.html) chosen for the build stage, even though it is never enabled/used?

I think we should go for the smallest one as long as it works. If that's possible, I guess it'd be nice to have a build without cds.

tech-6:
>* Why was a docker image with [Class Data Sharing](https://docs.oracle.com/en/java/javase/22/vm/class-data-sharing.html) chosen for the build stage, even though it is never enabled/used?

That's on me, Bellsoft's tag page is crazy full, will be pushing no cds on build stage shortly.

solonovamax:
although, using CDS/CRaC features could be something you look into, as it may improve performance.

CRaC would most likely need explicit support from reposilite, javalin, and ~~jetty~~ (jetty already seems to support it, with some additional configuration. see: [CRaC/docs/STEP-BY-STEP.md](https://github.com/CRaC/docs/blob/master/STEP-BY-STEP.md))

but CDS (& AppCDS/Dynamic CDS) could be a free performance boost at startup. (& help with memory? tbh idk exactly how CDS works, but the website I linked said it could help with memory) at the cost of a slightly larger docker image (tbh I wonder if the CDS archive files can be bundled into the jar, so that everyone can benefit from it)

dzikoysk:
I'd say this is completely out of scope for now to be fair - we should focus on providing smooth migration first.

It'd be great if you could quickly check if using these Temurin images would still require changes in permissions on our side. If no, then maybe we should use it for now. If we'd still need similar changes with Temurin tho, I guess we could proceed with Liberica for Reposilite 3.6.0, since you've already made these changes.

tech-6:
>If we'd still need similar (uid/gid) changes with Temurin tho, I guess we could proceed with Liberica for Reposilite 3.6.0, since you've already made these changes.

All Alpine based images will have a ping group for group 999.
```group
ping:x:999:
```
If we change to a ubuntu/debian based image such as `eclipse-temurin:21-jre-noble`, it would be larger but not have a group 999 conflict.

dzikoysk:
The slim image we're using right now is based on debian, so maybe let's do it the following way:

1. Move to temurin 21 based on debian in Reposilite 3.5.x
2. Extend our entrypoint file to provide a migration script that will start using a new default id.
  2.1. If existing data is owned by user with old default id, we'll also automatically try to update it to the new one. 
4. For next minor version (e.g. 3.6.0) we could migrate to liberica with a new setup based on alpine

tech-6:
Seems like a pretty good migration strategy, I'll write up changes to the scripts and make a PR, I'll rename this one and maybe return it to draft status.

solonovamax:
> I don't quite know how to benchmark the two images against each other. But if it offers great enough performance boost it could be worth keeping around.

afaik, CDS only affects memory for classes shared between multiple jvms, as well as decreases startup times. but I could be wrong.

tech-6:
> The slim image we're using right now is based on debian, so maybe let's do it the following way:
> 
> 1. Move to temurin 21 based on debian in Reposilite 3.5.x
> 2. Extend our entrypoint file to provide a migration script that will start using a new default id.
>    2.1. If existing data is owned by user with old default id, we'll also automatically try to update it to the new one.
> 3. For next minor version (e.g. 3.6.0) we could migrate to liberica with a new setup based on alpine

Changes made and PR created #2303

PR #2303, opened 2024-12-01
GH-2288 Migrate Docker images to Ubuntu and Eclipse Temurin and change user/group id

tech-6:
[As per discussed in #2289](https://github.com/dzikoysk/reposilite/pull/2289#issuecomment-2508117580)
- Update entrypoint.sh to be POSIX compatible (formerly used some BASH only syntax) for eventual switch to an Alpine Image.
- Update entrypoint.sh to include a warning about backwards compatibility and the changes coming to docker images with 3.6.X. Script will automatically change permissions of directories that require update.
- Use `eclipse-temurin:21-XXX-noble` images for build and run stages of Dockerfile.
- Speed up build time for local builds of Reposilite images by
  - Changing the ordering of Dockerfile instructions to allow for more parallel building and increased cache hits
  - Include an .dockerignore file to exclude most non-application files.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env sh
 
 # shellcheck disable=SC2086
 set -e
@@ -11,7 +11,8 @@ case "$REPOSILITE_OPTS" in
 esac
 
 # GH-1762: support for running as non-root user
-if (($(id -u) != 0)); then
+
+if [ "$(id -u)" != 0 ]; then
   exec java \
        -Dtinylog.writerFile.file="/var/log/reposilite/log_{date}.txt" \
        -Dtinylog.writerFile.latest=/var/log/reposilite/latest.log \
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -21,13 +21,26 @@ if [ "$(id -u)" != 0 ]; then
        $REPOSILITE_ARGS
 # GH-1200: run as non-root user
 else
+
+  # GH-2288: Dockerfile Step-1 migration warning
+  # shellcheck disable=SC2012
+  existing_uid=$(ls -dln data | awk '{print $3}')
+  # shellcheck disable=SC2012
+  existing_gid=$(ls -dln data | awk '{print $4}')
+  if [ $existing_uid != "977" ]||[ $existing_gid != "977" ]; then
+    printf "\033[1;31mStarting with Reposilite 3.5.20 the standard user id will be 977, I will make those changes now.\033[0m\n" 1>&2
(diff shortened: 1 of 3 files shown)


entrypoint.sh
```diff
@@ -20,13 +21,26 @@ if (($(id -u) != 0)); then
...
+
   # GH-1634: support custom user and group ids
-  GROUP_ID="${PGID:-999}"
+  GROUP_ID="${PGID:-977}"
   if ! grep -q "^reposilite" /etc/group;
   then
     addgroup --gid "$GROUP_ID" reposilite;
   fi
-  USER_ID="${PUID:-999}"
+  USER_ID="${PUID:-977}"
```
dzikoysk, line 43: I think that we should also include automatic permissions update, no? If we had an existing container with data being assigned to the user/group 999, then after running this one, it'll start using 977 and we'll run into permission issues :thinking: Group/user is not preserved in the volume, so we're pretty much creating it on each start in this case and that's why it may be an issue.

entrypoint.sh
```diff
@@ -20,13 +21,26 @@ if (($(id -u) != 0)); then
...
+  if [ $existing_uid != "977" ]||[ $existing_gid != "977" ]; then
+    printf "\033[1;31mStarting with Reposilite 3.5.20 the standard user id will be 977, I will make those changes now.\033[0m\n" 1>&2
+    printf "\033[1;31mAfter 3.6.0 docker installations with user id of 999 will no longer work.\033[0m\n" 1>&2
+    printf "\033[1;31mFor more information see: https://github.com/dzikoysk/reposilite/issues/2288\033[0m\n" 1>&2
+    printf "\033[1;31mIF YOU DOWNGRADE PAST THIS POINT \"Hic sunt dracones\"\033[0m\n" 1>&2
+  fi
+
   # GH-1634: support custom user and group ids
-  GROUP_ID="${PGID:-999}"
+  GROUP_ID="${PGID:-977}"
```
dzikoysk, line 38: By the way, any reason why it's 977? I don't mind, just being curious if you've noticed it's quite common value across other images, or it's just random :sweat_smile:

entrypoint.sh
tech-6: > I think that we should also include automatic permissions update, no?

Line 49-50 already does the permissions change that was needed. 

> Group/user is not preserved in the volume

It is preserved when using the same volume when using docker on a linux system and also internally when using Docker Desktop (MacOS and Windows use virtualized linux systems for docker containers)
image

entrypoint.sh
dzikoysk: Ok, let's keep it then :sweat_smile:

dzikoysk (approved):

dzikoysk:
I think it looks fine, thanks! I'll release it with Reposilite 3.5.20

Issue #2369, opened 2025-03-12
Option to require authentication to pull new artifacts from mirrors


kashike:
### Request details

When mirroring external repositories like Maven Central, it’s often desirable to mirror only the artifacts you need while avoiding unnecessary storage of artifacts required by external users.

In my case, I want artifacts to be pulled from the mirrors I’ve defined only when accessed by an authenticated user—not for anonymous access.

This is somewhat related to #1549, as both aim to reduce the number of stored artifacts.

PR #2370, opened 2025-03-18
Implement authenticated only fetching of uncached mirror artifacts

theminecoder:
Implements #2369 

This adds a new option to the maven mirrored repo settings that filters it out from the possible source of mirrors when there is no authenticated user in the request. Ideally this should only ever be used with the caching system turned on or another mirror host that has this option turned as otherwise this will deny/hide what used to be valid request.


dzikoysk:
Looks good I think :+1: I'm pretty busy this month, so thanks for covering it :heart_hands:

PR #2373, opened 2025-03-20
GH-2371 S3StorageProvider - fallback to AWS_REGION

### LINKED ISSUE #2371: S3 region - reposilite
### What happened?

Hi!

When I don't set region for the S3 storage provider explicitly, it falls back to resposilite. It looks like that was introduced in #1666 

When one runs in AWS, AWS normally injects the region via the environment variable AWS_REGION. For example when running in EKS.

From what I can see, you use environment variables only in the prometheus plugin, otherwise you always use the config. I would propose to do here something similar.

Current
```
        val region = when {
            settings.region.isNotEmpty() -> Region.of(settings.region)
            else -> Region.of("reposilite")
        }

        client.region(region)
```

Proposed

```
       // The AWS Java SDK uses by default the env variable AWS_REGION to determine the region
      // We use reposilite as a dummy region for S3-compatible providers, see https://github.com/dzikoysk/reposilite/issues/1666
        val region = when {
            settings.region.isNotEmpty() -> Region.of(settings.region)
            else -> System.getenv("AWS_REGION") ?: Region.of("reposilite")
        }

        client.region(region)
```

I am happy to submit a PR, I just wanted to check first if the solution is alright ;)

### Reposilite version

3.x

### Relevant log output

```shell

```

elruwen:
Resolves #2371

### DIFF
--- a/reposilite-backend/src/main/kotlin/com/reposilite/storage/s3/S3StorageProviderFactory.kt
+++ b/reposilite-backend/src/main/kotlin/com/reposilite/storage/s3/S3StorageProviderFactory.kt
@@ -63,10 +63,12 @@ class S3StorageProviderFactory : StorageProviderFactory<S3StorageProvider, S3Sto
             )
         }
 
-        val region = when {
-            settings.region.isNotEmpty() -> Region.of(settings.region)
-            else -> Region.of("reposilite")
-        }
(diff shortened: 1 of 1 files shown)


dzikoysk (approved):

dzikoysk:
Looks good, thanks!

PR #2375, opened 2025-03-20
GH-2374 Support AWS Web Identity Tokens

### LINKED ISSUE #2374: Support EKS Pod IAM role in S3 client config
### Request details

Hi!

I want to run reposilite inside EKS, which is the AWS provided Kubernetes solution. I want to use the S3 provider. So I am attaching an IAM role to the my pods.

In order to achieve that, the sts jar needs to be in the classpath, too.

See https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html

I will create a MR ;)

elruwen:
Resolves #2374

### DIFF
--- a/reposilite-backend/build.gradle.kts
+++ b/reposilite-backend/build.gradle.kts
@@ -76,6 +76,9 @@ dependencies {
     val awssdk = "2.30.36"
     implementation(platform("software.amazon.awssdk:bom:$awssdk"))
     implementation("software.amazon.awssdk:s3:$awssdk")
+    // STS is needed so it Web Identity Tokens can be used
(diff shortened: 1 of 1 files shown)


dzikoysk (approved):

dzikoysk:
Ugh, this lib has 0.5mb :roll_eyes: Anyway, looks good and it makes sense to include it, so thanks!

elruwen:
> Ugh, this lib has 0.5mb 🙄 Anyway, looks good and it makes sense to include it, so thanks!

Yeah :(
STS stands for security token service. Kinda scary how much code is needed for token management...

PR #2423, opened 2025-06-22
Emit ResolvedFileDataEvent for /{repository}/{gav}

IThundxr:


reposilite-backend/src/main/kotlin/com/reposilite/maven/RepositoryService.kt
```diff
@@ -162,6 +163,8 @@ internal class RepositoryService(
                 logger.debug("Cannot find '$gav' in '${repository.name}' repository, requesting proxied repositories")
                 mirrorService.findRemoteFile(repository, gav, accessToken)
             }
+        }.peek {
+            extensions.emitEvent(ResolvedFileDataEvent(accessToken, repository, gav, Result.ok(it)))
```
dzikoysk, line 167: `ResolvedFileDataEvent.result` is a mutable field, but when it's overridden by the listener, the new value is not used here. I think you should use `map` here on the emitted event.

dzikoysk (changes requested):

dzikoysk (approved):

dzikoysk:
Thanks! :heart_hands:

PR #2480, opened 2025-09-09
GH-2480 Fix vulnerability in rehype-prism (reposilite-site)

clintonsteiner:
* npm uninstall @mapbox/rehype-prism
* npm install rehype-prism-plus
* shift install to use new package


dzikoysk:
Oh, I didn't notice the old plugin is deprecated, thanks!

Issue #2558, opened 2026-01-31
Privacy Policy (Footer)


PinoOG:
### Request details

Hello, I already know it is possible to do it by deploying a static webpage, but I think, especially for EU laws, it would be good to have a small section in the admin dashboard to setup a simple privacy policy TEXT.

PR #2609, opened 2026-03-26
feat: Add support for legal notice in footer

DerFrZocker:
This PR, similar to the privacy policy #2558, ads a configurable link to a legal notice page, which in some countries is required.


dzikoysk:
Hmm I'm a bit afraid that it doesn't scale well. Maybe instead of adding new hardcoded types of links, we should rather move to a generic list, so users can add anything there at this point :thinking:

PR #2631, opened 2026-05-14
GH-2221 Resolution cache

### LINKED ISSUE #2221: Cache unresolvable artifacts too, like "Negative Cache"
### Request details

We got the ability to define the "max age of metadata file". 
This is extremely useful, if you do not want to send excessive amounts of requests to the upstream mirrors.

Will it be possible to add the ability to cache "unresolvable" artifacts too?
Nexus does give us this option, called "Negative Cache".

I understand, that this option may result in invalid responses, if something gets uploaded to a mirror - but the performance improvement otherwise should be massive.

dzikoysk:
Resolves #2221


vstrom650xt (approved):
123

Issue #2469, opened 2025-08-24
GitOps Suggestions


benalexau:
### Request details

Thanks for your work on Reposilite. I have recently deployed it into a Kubernetes cluster with Argo CD for a full GitOps-based workflow. This includes GitOps for deterministic, versioned user management and repositories. The `--shared-configuration` model was particularly useful for this, but there were a few areas I wanted to suggest in the hope they can be considered in future releases:

* Allow the `--token` CLI parameter to be provided via an environment variable. This would prevent the token password being visible in process lists. I currently work around this by deleting the temporary token at the end of a job that builds the users, but there is a temporary period where that temporary manager token is available. Most processes that require passwords allow them to be expressed as environment variables.
* Do not echo a token's password in the `token-generate` command response if that command specified the secret as an argument. Even if the command is sent over a `POST` to `/api/console/execute` it is still echoed to the JVM stdout and appears in logs. In GitOps environments (like Argo CD) the developers can typically see logs and it is undesirable this exposes the token secrets. To work around this I had to change the console logging level to `warn` which is pretty clunky (see code fragment below)
* Permit the user's routes to be included in `POST` calls to `/api/tokens/{name}`. This would avoid needing to use the `/api/console/execute` endpoint at all. Alternately, provide a different REST endpoint which can replace all of the user's routes. In either case please allow the JSON to fully replace (not simply augment) the token's routes.
* Support a `DELETE` to `/api/tokens/` which would delete all tokens except the manager token which was authenticated for that REST call. I currently work around this by listing all tokens and then deleting them one-by-one, excluding the temporary manager token. After creating the tokens (which includes a new manager) I then delete the temporary manager token.
[...]

PR #2649, opened 2026-06-03
GH-2469 Add option to not print regenerated secret

sebEg:
Analog to `token-generate`, add option to not print regenerated secret

#2469


dzikoysk (approved):

dzikoysk:
Thanks!

Issue #2650, opened 2026-06-03
Token update causes temporary 403 responses


sebEg:
### What happened?

Updating a token can cause brief 403 Forbidden responses for clients currently using that token.

It probably happens because the token is deleted and recreated during the update process, creating a short period where authentication fails.

This causes issues in our setup, where a job periodically updates tokens as part of GitOps reconciliation, even when the token configuration has not changed.

### Reposilite version

3.x

### Relevant log output

```shell

```

PR #2651, opened 2026-06-03
GH-2650 Avoid token deletion and recreation if it is of same type

sebEg:
Possible fix for #2650


dzikoysk (approved):

dzikoysk:
Thanks! I added some tiny adjustments to route & permission handling there too. In 4.x we should have a separate endpoint for updates :grimacing:

PR #2652, opened 2026-06-04
feat (frontend): Add internationalization support for multiple components and renderers

handy-git:
Add internationalization support for multiple components and renderers


dzikoysk:
Hey! As of right now internationalization is a little bit too much for our use-case and it's not something I want to include in 3.x, but feel free to open a new issue to track this as a feature request. In general, I recommend to always create a new issue first, so we can discuss this before preparing any implementation.

PR #2659, opened 2026-06-19
GH-2624 Fix usage calculations

### LINKED ISSUE #2624: Incorrect implementation of FS quotas

samfreund:
Previously, the check for usage on a given dir was done using the `getFileSize()` function. This function returned -1 when called on a directory, which was the case for any FileSystemStorageProvider. This meant that any checks against the size would always pass assuming the incoming file was less than the configured limit. 
This fix walks files to get the true size of the directory, caches the size, then checks incoming file(s) against that size, which should eventually block incoming requests, as opposed to the prior code, which would continue to accept requests until the disk was full (ask me how I know).

closes #2624

### DIFF
--- a/reposilite-backend/src/integration/kotlin/com/reposilite/storage/FileSystemStorageProviderIntegrationTest.kt
+++ b/reposilite-backend/src/integration/kotlin/com/reposilite/storage/FileSystemStorageProviderIntegrationTest.kt
@@ -23,6 +23,7 @@ import com.reposilite.status.FailureFacade
 import com.reposilite.storage.api.toLocation
 import com.reposilite.storage.filesystem.FileSystemStorageProvider
 import com.reposilite.storage.filesystem.FileSystemStorageProviderSettings
+import com.reposilite.storage.filesystem.NoQuota
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Nested
@@ -189,6 +190,43 @@ internal class FileSystemStorageProviderIntegrationTest : StorageProviderIntegra
                 }
             }
             error("Thread did not park or terminate, last state was ${thread.state}")
+        }
+
+    }
+
+    @Nested
+    inner class NoQuotaTests {
+
+        @BeforeEach
+        fun setup() {
+            val logger = InMemoryLogger()
+            val failureFacade = FailureFacade(logger)
+            val storageFacade = StorageFacade()
+
+            this@FileSystemStorageProviderIntegrationTest.storageProvider = storageFacade.createStorageProvider(
+                journalist = logger,
+                failureFacade = failureFacade,
+                workingDirectory = rootDirectory.toPath(),
+                repository = "test-noquota",
+                storageSettings = FileSystemStorageProviderSettings(quota = null)
+            )!!
+        }
+
+        @Test
+        fun `should report -1 usage when no quota is configured`() {
+            assertThat(assertOk(storageProvider.usage())).isEqualTo(-1L)
+        }
+
+        @Test
+        fun `should accept large upload when no quota is configured`() {
+            val payload = ByteArray(10 * 1024 * 1024) { 'b'.code.toByte() }
+            val response = storageProvider.putFile("/large.jar".toLocation(), payload.inputStream())
+            assertOk(response)
+        }
+
+        @Test
+        fun `should create NoQuota provider`() {
+            assertThat(storageProvider).isInstanceOf(NoQuota::class.java)
         }
 
     }
--- a/reposilite-backend/src/integration/kotlin/com/reposilite/storage/FileSystemStorageProviderIntegrationTest.kt
+++ b/reposilite-backend/src/integration/kotlin/com/reposilite/storage/FileSystemStorageProviderIntegrationTest.kt
@@ -167,9 +167,7 @@ internal class FileSystemStorageProviderIntegrationTest : StorageProviderIntegra
 
         @Test
         fun `should reject an upload meaningfully over the quota boundary`() {
-            // given: a payload comfortably above the configured 1 MB. Note: an exact `maxSize + 1`
-            // payload currently passes due to a one-byte slack inherited from `usage()` returning -1
-            // for the storage root — see docs/follow-ups/QUOTA_REDESIGN.md.
+            // given: a payload comfortably above the configured 1 MB
             val payload = ByteArray(2 * 1024 * 1024) { 'a'.code.toByte() }
 
             // when: the upload is attempted
--- a/reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
+++ b/reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
@@ -79,7 +79,7 @@ abstract class FileSystemStorageProvider protected constructor(
     val rootDirectory: Path,
 ) : StorageProvider {
 
-    private val usageCounterFile: Path = rootDirectory.resolve(".reposilite-usage")
+    private val usageCounterFile: Path = rootDirectory.resolve(".local/usage/${rootDirectory.name}")
     private val usageLock = Any()
     private var loadedUsage: Long = loadUsage()
 
--- a/reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
+++ b/reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
@@ -23,7 +23,6 @@ import com.reposilite.shared.notFound
 import com.reposilite.shared.stream.BoundedInputStream
 import com.reposilite.shared.stream.BoundedInputStreamLimitExceededException
 import com.reposilite.shared.toErrorResponse
-import com.reposilite.shared.internalServerError
 import com.reposilite.shared.toErrorResult
 import com.reposilite.storage.FilesComparator
 import com.reposilite.storage.StorageProvider
@@ -79,6 +78,43 @@ abstract class FileSystemStorageProvider protected constructor(
     val journalist: Journalist,
     val rootDirectory: Path,
 ) : StorageProvider {
+
+    private val usageCounterFile: Path = rootDirectory.resolve(".reposilite-usage")
+    private val usageLock = Any()
+    private var loadedUsage: Long = loadUsage()
+
+    private fun loadUsage(): Long =
+        try {
+            if (usageCounterFile.exists()) {
+                usageCounterFile.toFile().readText().trim().toLongOrNull()?.takeIf { it >= 0 }
+                    ?: recalculateUsage()
+            } else {
+                recalculateUsage()
+            }
+        } catch (e: Exception) {
+            recalculateUsage()
+        }
+
+    private fun recalculateUsage(): Long {
+        val size = Files.walk(rootDirectory).use { stream ->
+            stream.filter { it.type() == FILE }
+                .mapToLong { it.fileSize() }
+                .sum()
+        }
+        usageCounterFile.toFile().writeText("$size\n")
+        return size
+    }
+
+    private fun addToUsage(delta: Long) {
+        synchronized(usageLock) {
+            loadedUsage += delta
+            if (loadedUsage < 0) {
+                loadedUsage = recalculateUsage()
+            } else {
+                usageCounterFile.toFile().writeText("$loadedUsage\n")
+            }
+        }
+    }
 
     private enum class LockMode {
         READ,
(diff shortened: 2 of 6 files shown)


reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
```diff
@@ -78,6 +79,43 @@ abstract class FileSystemStorageProvider protected constructor(
     val rootDirectory: Path,
 ) : StorageProvider {
 
+    private val usageCounterFile: Path = rootDirectory.resolve(".reposilite-usage")
+    private val usageLock = Any()
+    private var loadedUsage: Long = loadUsage()
```
dzikoysk: I don't think it makes sense to track usage on all repositories with this system, so let's make it nullable and disable quota for new repositories as a default if not provided. It might be also a good idea to wrap it into some dedicated class like `UsageTracker` to simplify handling.

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
```diff
@@ -78,6 +79,43 @@ abstract class FileSystemStorageProvider protected constructor(
     val rootDirectory: Path,
 ) : StorageProvider {
 
+    private val usageCounterFile: Path = rootDirectory.resolve(".reposilite-usage")
```
dzikoysk: Keeping it in the root directory would leak this file to all users. We could try to filter this out from the results, but then it feels a little bit dirty. Maybe we should keep this like javadocs, outside of the repository, maybe in `.local/usage/<repository-name>` for now. We already keep `.local/reposilite.address` there and this is just yet another information about the local state that we can recalculate on startup.

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
```diff
@@ -147,9 +185,11 @@ abstract class FileSystemStorageProvider protected constructor(
                             // Re-check post-write: InputStream.available() is just a hint and returns 0 for HTTP request bodies.
                             .flatMap { canHold(temporaryFile.fileSize()).mapErr { INSUFFICIENT_STORAGE.toErrorResponse("Not enough storage space available: ${it.message}") } }
                             .peek {
+                                val size = temporaryFile.fileSize()
                                 acquireFileAccessLock(location, LockMode.WRITE).use {
                                     temporaryFile.moveTo(file, overwrite = true)
                                 }
+                                addToUsage(size)
```
dzikoysk: Overriding a file creates a false positive usage allocation, whenever usage is enabled we need to also check if file already exists and insert the delta there - the file might be even smaller, so in such case this would effectively remove usage.

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
```diff
@@ -78,6 +79,43 @@ abstract class FileSystemStorageProvider protected constructor(
...
+                .mapToLong { it.fileSize() }
+                .sum()
+        }
+        usageCounterFile.toFile().writeText("$size\n")
+        return size
+    }
+
+    private fun addToUsage(delta: Long) {
+        synchronized(usageLock) {
+            loadedUsage += delta
```
dzikoysk: I think it breaks multi-instance setup, to make sure it works we would need to read it on each write :thinking: Of course there is still chance to miss something, but I guess we don't really need to be strictly precise with it in such rare case.

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
```diff
@@ -288,7 +337,7 @@ abstract class FileSystemStorageProvider protected constructor(
             .fold({ true }, { false })
 
     override fun usage(): Result<Long, ErrorResponse> =
-        getFileSize(Location.empty())
+        loadedUsage.asSuccess()
```
dzikoysk: When quota is disabled, I think it would be fine to return `null` as success

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
dzikoysk: `loadedUsage` does not have synchronized reads and it's not even marked as volatile, so you can get stale value in each thread. I guess it could only bite us at the end of the capacity, so a few files could exceed the quota, low risk but might be nice to do it properly.

reposilite-backend/src/integration/kotlin/com/reposilite/storage/FileSystemStorageProviderIntegrationTest.kt
dzikoysk, line 1: The test suite is quite poor in that area, it would be nice to cover cases I mentioned above like repeated overrides, overriding a bigger file with a smaller one, maybe something else you could think of :thinking:

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
```diff
@@ -78,6 +79,43 @@ abstract class FileSystemStorageProvider protected constructor(
...
+        try {
+            if (usageCounterFile.exists()) {
+                usageCounterFile.toFile().readText().trim().toLongOrNull()?.takeIf { it >= 0 }
+                    ?: recalculateUsage()
+            } else {
+                recalculateUsage()
+            }
+        } catch (e: Exception) {
+            recalculateUsage()
+        }
```
dzikoysk: It's probably better to not start at all in such case, assuming that we will have this `UsageTracker`, you could try to create it in the factory and if it fails, we should communicate error and skip given repository.

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
```diff
@@ -78,6 +79,43 @@ abstract class FileSystemStorageProvider protected constructor(
...
+        usageCounterFile.toFile().writeText("$size\n")
+        return size
+    }
+
+    private fun addToUsage(delta: Long) {
+        synchronized(usageLock) {
+            loadedUsage += delta
+            if (loadedUsage < 0) {
+                loadedUsage = recalculateUsage()
```
dzikoysk: Do you see a case when it can happen? :eyes:

samfreund: It's not horribly likely, but if a file is modified by something other than reposilite and increases in size, then gets deleted, the usage could dip below 0. Upon thinking about it, I don't think this would realistically get hit. A better choice would probably be recalculating usage once a week (or something like that).

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
samfreund: I did Long.MAX_VALUE, that makes more sense to me as it represents an "infinite" quota. I can do null instead tho if you'd like.

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
samfreund: By skip, do you mean skip accepting files, or just set the quota to None.

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/UsageTracker.kt
```diff
@@ -0,0 +1,56 @@
...
+    fun getUsage(): Long = loadedUsage
+
+    fun addDelta(delta: Long) {
+        synchronized(lock) {
+            loadedUsage = loadUsage() + delta
+            if (loadedUsage < 0) {
+                loadedUsage = recalculate()
+            } else {
+                counterFile.createParentDirectories()
+                counterFile.toFile().writeText("$loadedUsage\n")
```
mcm001, line 29: Does text or binary make more sense for this file? I don't think either one particularly matters, just curious to hear thoughts

samfreund: text is nice if you want to verify manually

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/UsageTracker.kt
```diff
@@ -0,0 +1,56 @@
...
+import java.nio.file.Path
+import kotlin.io.path.createParentDirectories
+import kotlin.io.path.exists
+import kotlin.io.path.fileSize
+import kotlin.io.path.name
+
+internal class UsageTracker(
+    private val rootDirectory: Path,
+) {
+    private val counterFile: Path = Path.of("/app/data").resolve(".local/usage/${rootDirectory.name}")
```
dzikoysk, line 15: `Path.of("/app/data")` - this is incorrect, you should working directory that can be obtained from `ReposiliteParameters`. Every plugin has access to it via `parameters().workingDirectory,`

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/UsageTracker.kt
dzikoysk: I'd stay with text, it's better to use human-readable formats whenever possible.

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
dzikoysk: By skipping, I mean to not load repository at all in the first place - take a look at `RepositoryProvider.createRepositories`, whenever there is a problem with configuration, we do not load it to make sure that corrupted repository cannot be used by users.

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
dzikoysk: Hmm let's maybe make sure that when you write the usage to file, you filter out such cases by using something like `Math.max(0, updatedUsage)`. I think it's usually better to make sure that the state is in correct shape rather than treating further side-effects.

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
dzikoysk: Our S3 implementation returns -1 in such case:

```
override fun usage(): Result<Long, ErrorResponse> = ok(-1)
```

But to be fair, I don't quite like magic numbers that has some implicit meaning. I think it's way better to represent it at signature level and be honest about the actual result of the function - if someone would use this API in the future, I'm quite sure that they wouldn't even notice that in such case we don't provide real usage.  Let's align both places to the null value :eyes:

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
```diff
@@ -250,10 +255,19 @@ abstract class FileSystemStorageProvider protected constructor(
...
+                            .mapToLong { it.fileSize() }
+                            .sum()
+                    }
+                    else -> rootPath.fileSize()
+                }
                 when {
                     rootPath.isDirectory() -> rootPath.deleteRecursively()
                     else -> rootPath.deleteExisting()
                 }
+                usageTracker?.addDelta(-removedSize)
```
dzikoysk, line 270: The entire repeated file walk is now executed even if tracker is not enabled. Maybe it'd be a good idea to add utility functions similar to `rootPath.deleteRecursively()/deleteExisting()` that returns usage (or null if failed, as a replacement to boolean true/false result)?

reposilite-backend/src/main/kotlin/com/reposilite/storage/filesystem/FileSystemStorageProvider.kt
```diff
@@ -73,9 +74,10 @@ import kotlin.io.path.useDirectoryEntries
 /**
  * @param rootDirectory root directory of storage space
  */
-abstract class FileSystemStorageProvider protected constructor(
+abstract class FileSystemStorageProvider internal constructor(
```
dzikoysk, line 77: Is this needed? :thinking:

PR #2664, opened 2026-06-30
GH-2663 Honor X-Forwarded-Prefix

### LINKED ISSUE #2663: Index links drop the reverse-proxy path prefix (X-Forwarded-Prefix is ignored)
### What happened?

When Reposilite is served behind a reverse proxy under a sub-path, the HTML
repository index pages generate links that drop the proxy's path prefix, so
every link on the page is broken.

**Setup**

The proxy exposes Reposilite under a `/maven` sub-path and strips that prefix
before forwarding (standard reverse-proxy mounting):

- Public URL:  `https://repo.example.com/maven/public-release/cz/tacr/elza/elza-distribution/3.3.16/`
- Internal URL: `https://internal.server:8084/public-release/cz/tacr/elza/elza-distribution/3.3.16/`

(`public-release` is the repository name; `/maven` is the proxy mount prefix.)

**Steps to reproduce**

1. Run Reposilite behind a reverse proxy mounted on a sub-path (`/maven`)
   that rewrites/strips the prefix before forwarding to the backend.
2. In a browser, open a repository directory through the public URL, e.g.
   `https://repo.example.com/maven/public-release/cz/tacr/elza/elza-distribution/3.3.16/`.

**Expected**

Links to the listed files and the parent directory keep the `/maven` prefix, e.g.
`https://repo.example.com/maven/public-release/.../3.3.16/elza-distribution-3.3.16.pom`.

**Actual**

The page is rendered with `<base href='/public-release/.../3.3.16/'>` — an absolute
path that does not include `/maven`. The browser therefore resolves every child and
parent link to `https://repo.example.com/public-release/...`, dropping `/maven`, and
all links 404 (or escape the proxy mount entirely).

**Cause**

Reposilite only knows its configured `basePath` (default `/`) and ignores the
`X-Forwarded-Prefix` header. Because the proxy strips the prefix, the backend has
no way to learn it, so the server-rendered directory index
(`createDirectoryIndexPage`) and the 404 page emit URLs without `/maven`.

The existing static `basePath` configuration is not a good fit here: it would have to
be hardcoded to `/maven`, which then breaks direct/internal access to the same
instance and cannot serve multiple prefixes.

**Proposed fix**

Honor the de-facto-standard `X-Forwarded-Prefix` header per request when generating
links on the repository index and 404 pages, falling back to the configured `basePath`
when the header is absent (so current behavior is unchanged). The header value is
client-controlled and would be validated/sanitized before use. I'll open a PR
implementing this.

### Reposilite version

3.x

### Relevant log output

```shell

```

PetrPytelka:
Resolves #2663

## Problem

When Reposilite is served behind a reverse proxy mounted on a sub-path (e.g.
`/maven`) that strips the prefix before forwarding, the server-rendered
repository index pages emit `<base href>` and links as absolute paths without
that prefix. The browser then resolves every child/parent link outside the
proxy mount, so all links on the page are broken.

Reposilite only knew its statically configured `basePath` (default `/`) and
ignored the `X-Forwarded-Prefix` header, so it had no way to learn the prefix
the proxy had stripped.

## Solution

Honor the de-facto-standard `X-Forwarded-Prefix` header per request when
generating links, falling back to the configured `basePath` when the header is
absent or invalid. The same instance therefore works correctly both directly
and behind a prefix-stripping proxy, without hardcoding `basePath`.

The header is client-controlled, so its value is strictly validated and
normalized before being placed into any HTML attribute.

## Changes

- `BasePathFormatter.formatForwardedBasePath()` — validates the header against a
  whitelist (`/?[a-zA-Z0-9._~/-]*`, rejects `//`) and normalizes it, returning
  `null` when unsafe.
- `FrontendFacade.resolveBasePath(prefix)` — resolves the effective base path
  per request (header → else configured `basePath`); adds the
  `X_FORWARDED_PREFIX` constant.
- `MavenEndpoints` — the directory index page and the 404 page now use the
  per-request base path.
- `NotFoundHandler` — passes the header through to the 404 page.
- Docs: nginx guide gains a short note on the `X-Forwarded-Prefix` option.

## Security

The forwarded prefix ends up inside HTML attributes (`<base href>`, `<a href>`),
so it is whitelisted to a simple path and `//` is rejected, preventing
attribute-breakout XSS and protocol-relative redirect hijack (`//evil.com`).
Invalid values silently fall back to the configured `basePath`. The directory
index is not cached, so a crafted header only affects the attacker's own
response.

## Backward compatibility

No new configuration and no behavioral change when the header is absent:
`resolveBasePath` falls back to the existing `basePath`, so default and
custom-`basePath` deployments are unaffected.

## Testing

- New `BasePathFormatterTest` covers normalization of valid prefixes and
  rejection of unsafe ones.
- `./gradlew :reposilite-backend:test` passes; backend compiles cleanly.

## Out of scope

The SPA dashboard still uses the configured `basePath` placeholder; this PR
targets the repository content (index/404) pages, which is where broken links
are observed. Honoring the prefix in the dashboard can follow separately.


dzikoysk:
Thanks for the PR, it looks like AI generated content tho, so I started working on this, I hope you don't mind. 

For backend related stuff it works... but not really for the entire frontend that is still filled with the preconfigured base path. I think we should actually make it work for both if we want to support it - I think it should be opt-in config option too, otherwise it could be a breaking change for some users.

PR #2680, opened 2026-07-12
GH-2680 Fix link for installation not going to the propper place

randomusert:
If you are already in /guide it will go to /guide/guide/general for example. This behaviour was caused by the link being guide/general not /guide/general which means that it would link to the incorrect place


dzikoysk:
Oh I missed that, thanks!

PR #2697, opened 2026-08-05
Fix wrong sort in Extensions.registerEvent (Fixes #2696)

### LINKED ISSUE #2696: Event handler priorities not working because of wrong sorting method
### What happened?

List.sortedBy needs to be replaced with MutableList.sortBy so that the list is actually mutated in-place when registering events with non-default priority.

on this line: https://github.com/dzikoysk/reposilite/blob/161d9ae817ef71a433e24ee62cd8f963ada132b8/reposilite-backend/src/main/kotlin/com/reposilite/plugin/Extensions.kt#L62

i noticed this while i was trying to make a plugin that registers http handler events before the default frontend to override parts of the frontend without fully disabling it,

### Reposilite version

3.x

### Relevant log output

```shell

```

FalsePattern:


dzikoysk:
Thanks!

Issue #2056, opened 2024-02-26
Reposilite should search all mirrors when looking up the last version of a package


Jean-Daniel:
### Request details

I use to have an artifactory virtual repository exposing both the Google maven repo (https://dl.google.com/android/maven2/) and maven central (https://repo1.maven.org/maven2/).

Unfortunately, both repositories exposes a bunch of deprecated packages, that use to be published in one repo, and are now published in the other.

When requesting the last version of a package like `com.android.tools.build:gradle`, if reposilite lookup maven central first, it find the version 2.3.3, and does no try to search for more recent version in the Google repo.

On the other side, when looking up for `com.google.devtools.ksp:symbol-processing`, this is the Google repo that return an old release.

Artifactory was handling it properly by returning the most recent version for both packages.

An other way to help mitigation this issue, would be to support a `groups blacklist` in the mirror setting to be able to tell reposilite to ignore some mirror for a specific package.

dzikoysk:
Yeah, I think this kind of enhancement makes sense for such a weird scenarios. For now, I think you could try to use this functionality:

* https://reposilite.com/guide/mirrors#allowed-groups

to only list these particular artifact groups for each repo. A bit annoying, but should work as a quick solution.

ginkel:
> Any plans to provide a groups denylist? This would be helpful in order to mitigate dependency confusion attacks.

I have implemented a group blocklist and am preparing a PR.

PR #2717, opened 2026-08-22
GH-2056 Add blockedGroups to mirrored repository settings

ginkel:
…2056)

Mirror resolution returns the first host that answers, in configured order, so a group carried by several mirrors is served by whichever comes first - returning a stale version, or letting a public mirror answer for an internal group ahead of the mirror that should serve it.

allowedGroups cannot correct this: confining a mirror to the groups it should answer for means enumerating every other group it must still serve. blockedGroups inverts that, and is checked first so it takes precedence.

Matching reuses the existing raw path-prefix test. Blank entries are ignored. Only the mirror leg is suppressed - artifacts in local storage still resolve.

### DIFF
--- a/reposilite-backend/src/main/kotlin/com/reposilite/maven/MirrorService.kt
+++ b/reposilite-backend/src/main/kotlin/com/reposilite/maven/MirrorService.kt
@@ -204,7 +204,8 @@ internal class MirrorService(
 
     private enum class DisallowedReason {
         EXTENSION,
-        GROUP
+        GROUP,
+        BLOCKED_GROUP
     }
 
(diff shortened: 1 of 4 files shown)


dzikoysk (approved):

dzikoysk:
Thanks! 

In general it looks good, I noticed that with the resolution cache enabled, an inherited mirror pin is applied before mirror eligibility is evaluated (so if the pinned mirror blocks a descendant path, resolution returns `NoEligibleHosts` instead of trying another configured mirror)... :/ But it predates this PR and can also affect existing eligibility filters, so I’ll try to tackle this separately on top of these changes.

